PR #5, opened 2018-05-07
Don't give a null serverConnection to DnsResolver

bemasc:
DnsResolverUdpToHttps can't possibly work if serverConnection is null.
This may only happen in situations that wouldn't work anyway
(e.g. bootstrap failed), but it's still illogical and is producing
a lot of error reports.

Change-Id: I30f33209bf32d56060e6a3d92ab70184bd12e282

### DIFF
--- a/Android/app/src/main/java/app/intra/DnsVpnService.java
+++ b/Android/app/src/main/java/app/intra/DnsVpnService.java
@@ -286,14 +286,14 @@ public class DnsVpnService extends VpnService implements NetworkManager.NetworkL
     }
   }
 
-  private void startDnsResolver() {
-    if (dnsResolver == null) {
+  private synchronized void startDnsResolver() {
+    if (dnsResolver == null && serverConnection != null) {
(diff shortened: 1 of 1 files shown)


Android/app/src/main/java/app/intra/DnsVpnService.java
```diff
@@ -286,14 +286,14 @@ private void pingLocalDns() {
     }
   }
 
-  private void startDnsResolver() {
-    if (dnsResolver == null) {
+  private synchronized void startDnsResolver() {
+    if (dnsResolver == null && serverConnection != null) {
```
alalamav: Should we submit a crash report if `serverConnection == null`?

bemasc: I don't think it's necessarily an error.  For example, if `onNetworkConnected` runs during bootstrap, and bootstrap fails, then `onNetworkConnected` will call `startDnsResolver`, but `dnsResolver` and `serverConnection` will both be `null`.

It does seem like there's probably a better way, but fixing it might require rethinking the whole activation user experience.

PR #7, opened 2018-05-09
Korean trasnlation

alexkoala-old:


alexkoala:
I signed it!

cjhenck:
Thanks for your contribution! The user interface isn't standardized yet so we aren't ready to add more languages, but we will certainly incorporate your translation once we're ready.

May I ask how you found out about Intra? Is it working well for you?

alexkoala:
I found this app in korea wiki called namu.wiki. [Here is the link.](https://namu.wiki/w/%EB%B0%A4%ED%86%A0%EB%81%BC)
The South Korea goverment has started DNS censorship. If someone lives in korea access the blocked website using http, it redirects to [warning.or.kr](http://warning.or.kr) because Korea goverment is doing DPI. That means If server uses https, Korea goverment can't block website. The example is [4shared.com](https://www.4shared.com/). But Korea Goverment started DNS censorship in May 4, 2018. It blocks the user who wants to access blocked website using https. I think this way is very wrong because Korea goverment can abuse this technology.
Of course people can just change their DNS to Google Public DNS or Cloudflare DNS, but it is annoying. This app is really easy to use. I wanted to help this project to many korean can use this app.

And the app works very good.

PR #8, opened 2018-05-17
Minor crash reporting improvements

bemasc:


Android/app/src/main/java/app/intra/DnsVpnService.java
```diff
@@ -280,9 +280,10 @@ private void pingLocalDns() {
               zero, zero.length, 0, InetAddress.getByName(privateIpv4Address.router), 53);
       DatagramSocket datagramSocket = new DatagramSocket();
       datagramSocket.send(packet);
-    } catch (Exception e) {
-      FirebaseCrash.logcat(Log.ERROR, LOG_TAG, "Failed to send UDP ping.");
-      FirebaseCrash.report(e);
+    } catch (IOException e) {
```
alalamav: Can we get other exceptions? Should we report those?

bemasc: If there were other kinds of exceptions, this code wouldn't compile in Java ... except that this doesn't apply to runtime exceptions.  I think the only exceptions we've seen here have been IOException, so hopefully this is safe.  Usually in Java it's recommended to catch the narrowest Exception type possible.

PR #12, opened 2018-06-12
Centralize preference state

bemasc:
In addition to improving code health, this removes the VPN
service's reliance on Intent extras, which is a prerequisite
for always-on/autostart.

### DIFF
--- a/Android/app/src/main/java/app/intra/Preferences.java
+++ b/Android/app/src/main/java/app/intra/Preferences.java
@@ -1,0 +1,106 @@
+package app.intra;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Static class representing on-disk storage of mutable state.  Collecting this all in one class
+ * helps to reduce duplication of code using SharedPreferences, allows settings to be read and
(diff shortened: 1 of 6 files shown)


Android/app/src/main/java/app/intra/Preferences.java
```diff
@@ -0,0 +1,106 @@
...
+  // TODO: Unify preferences into a single file.
+  private static final String APPROVAL_PREFS_NAME = "IntroState";
+
+  private static SharedPreferences getSettings(Context context) {
+    return context.getSharedPreferences(MAIN_PREFS_NAME, Context.MODE_PRIVATE);
+  }
+
+  public static String getServerName(Context context) {
+    SharedPreferences settings = getSettings(context);
+
```
alalamav: nit: whitespace after `getSettings` in this file

Android/app/src/main/java/app/intra/Preferences.java
```diff
@@ -0,0 +1,106 @@
...
+    String defaultDomain = context.getResources().getStringArray(R.array.domains)[0];
+    return settings.getString(SERVER_KEY, defaultDomain);
+  }
+
+  public static boolean setServerName(Context context, String name) {
+    String oldName = getServerName(context);
+    if (name.equals(oldName)) {
```
alalamav: Can `name` be null?

Issue #11, opened 2018-06-10
Always On VPN


hpfm-2:
Can the app support Always on VPN and/or ability to turn on at startup?

andrebrait:
Besides the starting on boot up, the rest is already possible, I think. Android has built in a setting for always-on VPN and mine has been running without disconnects for days now.

PR #14, opened 2018-06-13
Add autostart-on-boot

bemasc:
This change relaunches Intra on boot if it was active at shutdown,
so that the VPN stays active until the user turns it off.

This is relevant to #11.

### DIFF
--- a/Android/app/src/main/java/app/intra/AutoStarter.java
+++ b/Android/app/src/main/java/app/intra/AutoStarter.java
@@ -1,0 +1,38 @@
+package app.intra;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.net.VpnService;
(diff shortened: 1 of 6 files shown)


Android/app/src/main/java/app/intra/AutoStarter.java
```diff
@@ -0,0 +1,38 @@
...
+import android.content.Context;
+import android.content.Intent;
+import android.net.VpnService;
+import android.util.Log;
+import com.google.firebase.crash.FirebaseCrash;
+
+/**
+ * Broadcast receiver that runs on boot.
+ */
+
```
alalamav: nit: this also runs when the app restarted due to an update; extra newline here.

bemasc: Fixed.

PR #24, opened 2018-07-27
Allow apps to bypass the VPN

### LINKED ISSUE #18: Intra interferes with WebRTC
Intra seems to interfere with Threema's WebRTC connection to Threema Web 

possibly relevant:  https://github.com/julian-klode/dns66/issues/193

bemasc:
In practice, apps could easily bypass Intra anyway, e.g. by
sending queries to a hardcoded resolver.  Explicitly allowing
bypass fixes #18 (interference with some WebRTC-based apps).

### DIFF
--- a/Android/app/src/main/java/app/intra/DnsVpnService.java
+++ b/Android/app/src/main/java/app/intra/DnsVpnService.java
@@ -436,7 +436,11 @@ public class DnsVpnService extends VpnService implements NetworkManager.NetworkL
               .addDnsServer(privateIpv6Address.router);
 
       if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-        builder = builder.setBlocking(true); // Only available in API >= 21
+        // Only available in API >= 21
+        builder = builder.setBlocking(true);
+        // Some WebRTC apps rely on the ability to bind to specific interfaces, which is only
+        // possible if we allow bypass.
+        builder = builder.allowBypass();
 
         try {
           // Workaround for any app incompatibility bugs.


cjhenck:
LGTM

Android/app/src/main/java/app/intra/DnsVpnService.java
```diff
@@ -436,7 +436,11 @@ private ParcelFileDescriptor establishVpn() {
               .addDnsServer(privateIpv6Address.router);
 
       if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-        builder = builder.setBlocking(true); // Only available in API >= 21
+        // Only available in API >= 21
+        builder = builder.setBlocking(true);
+        // Some WebRTC apps rely on the ability to bind to specific interfaces, which is only
+        // possible if we allow bypass.
+        builder = builder.allowBypass();
```
ignoramous, line 443: Isn't [`allowBypass` required only for apps that don't offer full routing](https://cs.android.com/android/platform/superproject/+/android-r-beta-3:frameworks/base/services/core/java/com/android/server/ConnectivityService.java;bpv=0;bpt=0;l=6406) (unlike Intra which does route _all_ of the IPv4 space)?

I read the source and I see either [`RULE_PRIORITY_SECURE_VPN` or `RULE_PRIORITY_SECURE_VPN` `fwmark`s set on `netId`](https://cs.android.com/android/platform/superproject/+/android-r-beta-3:system/netd/server/RouteController.cpp;l=815;drc=android-r-beta-3) and `socket`s depending on if the VPN is `secure` (which it is when `allowBypass` is _not_ set **or** the VPN is in "lockdown mode" aka the "Block connections without VPN" mode settable by the user through Android's VPN settings), which (I am not sure) are used to make routing decisions later? But it is unclear to me if webrtc initiated ICE session benefit from this? Is there a way to test this?).

An unintended effect of setting `allowBypass` is the VPN is without Internet connectivity if the user also chooses to start it in "lockdown mode": [Ref](https://github.com/celzero/rethink-app/issues/92); and so, `allowBypass`  must be unset when the VPN is locked-down, like so [celzero/rethink-app@71e1a840b](https://github.com/celzero/rethink-app/commit/71e1a840b5eb5c865081bd883019bcec7f433551), to unblock connectivity to the Internet.

ignoramous: > Isn't [`allowBypass` required only for apps that don't offer full routing](https://cs.android.com/android/platform/superproject/+/android-r-beta-3:frameworks/base/services/core/java/com/android/server/ConnectivityService.java;bpv=0;bpt=0;l=6406) (unlike Intra which does route all of the IPv4 space)?

In addition to this, it might be worth taking a look at [`IntraVpnService#setUnderlyingNetworks`](https://github.com/Jigsaw-Code/Intra/blob/27b8befabe/Android/app/src/main/java/app/intra/sys/IntraVpnService.java) which only adds one network, the current active-network, but could also add, in addition, all other networks [the OS is connected to](https://developer.android.com/reference/android/net/ConnectivityManager#getAllNetworks()) (with active-network in he first position in the array).

Also: `allowBypass`, from the documentation, lets an app bind its sockets to any network, but if all networks are also bound to VPN, then they might not need a bypass in the first place?

Issue #19, opened 2018-06-21
Some downloads don't work when Intra is active


bemasc:
Downloads that are implemented using DownloadManager and JobScheduler often don't work through Intra.  This appears to be because Intra's VPN, with its narrow routing, is not marked by Android as having the INTERNET capability.  (The Android source code indicates that a VPN must include 85% of the routing space to qualify for this mark.)  When the default network doesn't have the INTERNET capability mark, DownloadManager hangs (waiting for internet access).

It's not yet clear how we can mitigate this problem in Intra.

bemasc:
It seems like this problem only affects Android 7.0 - 8.1, and is related to the VPN being marked as not having the "INTERNET" attribute, because it only routes a small amount of the address space.

PR #25, opened 2018-07-27
Enable app bypass

bemasc:
This allows a user to select certain apps that should not be
routed through Intra.

This is a workaround for #19, but may also be useful for
advanced users in other situations.

### DIFF
--- a/Android/app/src/main/java/app/intra/SettingsFragment.java
+++ b/Android/app/src/main/java/app/intra/SettingsFragment.java
@@ -81,6 +81,10 @@ public class SettingsFragment extends PreferenceFragmentCompat {
         // Showing ourselves in the list of apps would be confusing and useless.
         continue;
       }
+      if (pm.getLaunchIntentForPackage(info.packageName) == null) {
+        // This is not an app that can actually be launched, so it's confusing to show it.
+        continue;
+      }
       entries.add(new Entry(info.loadLabel(pm).toString(), info.packageName));
     }
     Collections.sort(entries);
--- a/Android/app/src/main/java/app/intra/SettingsFragment.java
+++ b/Android/app/src/main/java/app/intra/SettingsFragment.java
@@ -1,0 +1,94 @@
+package app.intra;
+
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.os.Build;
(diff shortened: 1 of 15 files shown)


Android/app/src/main/java/app/intra/SettingsFragment.java
```diff
@@ -0,0 +1,94 @@
...
+        // Showing ourselves in the list of apps would be confusing and useless.
+        continue;
+      }
+      entries.add(new Entry(info.loadLabel(pm).toString(), info.packageName));
+    }
+    Collections.sort(entries);
+    labels = new ArrayList<>(entries.size());
+    packageNames = new ArrayList<>(entries.size());
+    for (Entry e : entries) {
+      labels.add(e.label);
```
cjhenck: Should we use the package name as a backup label?

Android/app/src/main/java/app/intra/SettingsFragment.java
```diff
@@ -0,0 +1,94 @@
...
+      apps.setSummary(R.string.old_android);
+    }
+  }
+
+  /**
+   * Update this SettingsFragment's list of installed apps for the "Excluded apps" setting.  This
+   * method must be called before the fragment is displayed.
+   * @param pm The global PackageManager.
+   */
+  public void updateInstalledApps(PackageManager pm) {
```
cjhenck: Do you think all installed packages is the most sensible?

I've been debating whether some combo of [LauncherApps](https://developer.android.com/reference/android/content/pm/LauncherApps) and system packages might be a better UX, but I think we wouldn't hit our primary use case (download manager)

bemasc: Alas, the download manager runs as the app that is using it, so AFAIK there's no way to blacklist it universally.  I did try!

This code uses [`getInstalledApplications`](https://developer.android.com/reference/android/content/pm/PackageManager#getInstalledApplications(int)), not `getInstalledPackages`, so it only collects the "application packages", not "all installed packages".  However, it does seem to be pulling in a lot of packages that aren't normally user-visible.  I added a check to exclude non-launchable packages, which should make the list shorter and more comprehensible.

bemasc:
@cjhenck Please re-review.

PR #26, opened 2018-07-31
Move server selection to Settings and allows URLs

### LINKED ISSUE #21: Would be nice to have custome dns over https input field
I have my own doh server, so would be nice to have this option

bemasc:
With this change, the user can select a server before initiating a
connection for the first time (important if the default server
is not working for any reason).  This change also allows advanced
users to use arbitrary DOH servers, not just the two presets.

This change removes the server control from the main screen.
This is good for visual consistency, but if users find it harder
to discover this option then that could be an unfortunate side-effect.

Fixes #21


Android/app/src/main/java/app/intra/ServerChooserFragment.java
```diff
@@ -0,0 +1,155 @@
...
+        bundle.putString(ARG_KEY, key);
+        fragment.setArguments(bundle);
+        return fragment;
+    }
+
+    @Override
+    public void onCheckedChanged(RadioGroup group, int checkedId) {
+        if (checkedId == R.id.pref_server_google) {
+            newUrl = getResources().getString(R.string.url0);
+            setValid(true);
```
cjhenck: I personally like to set the UI based on state.  E.g. update newURL, then call an updateUI() function, or have a setter of some sort for newURL. Makes it easier to maintain the UI logic.

Android/app/src/main/java/app/intra/ServerChooserFragment.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Override
+    public void onTextChanged(CharSequence s, int start, int before, int count) {
+    }
+
+    @Override
+    public void afterTextChanged(Editable s) {
+        newUrl = s.toString();
+        setValid(checkUrl(Untemplate.strip(newUrl)));
```
cjhenck: E.g. this would end up in updateUI as well

Android/app/src/main/java/app/intra/ServerChooserFragment.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // Check that the URL is a plausible DOH server: https with a domain, a path (at least "/"),
+    // and no query parameters or fragment.
+    private boolean checkUrl(String url) {
+        try {
+            URL parsed = new URL(url);
+            return parsed.getProtocol().equals("https") && !parsed.getHost().isEmpty() &&
+                !parsed.getPath().isEmpty() && parsed.getQuery() == null && parsed.getRef() == null;
```
cjhenck: Do you think we really need to have a path?  People type in google.com [Enter] all the time?

Android/app/src/main/java/app/intra/ServerChooserFragment.java
```diff
@@ -0,0 +1,155 @@
...
+    @Override
+    protected void onBindDialogView(View view) {
+        super.onBindDialogView(view);
+        ServerChooser preference = (ServerChooser) getPreference();
+        String url = preference.getUrl();
+        buttons = view.findViewById(R.id.pref_server_radio_group);
+        text = view.findViewById(R.id.custom_server_url);
+        if (url == null || url.equals(getResources().getString(R.string.url0))) {
+            buttons.check(R.id.pref_server_google);
+            text.setEnabled(false);
```
cjhenck: And this

Android/app/src/main/java/app/intra/MainActivity.java
```diff
@@ -134,13 +134,27 @@ private void openUrl(String url) {
...
     super.onCreate(savedInstanceState);
 
+    // Sync old settings into new preferences if necessary.
+    PersistentState.syncLegacyState(this);
+
     // Export defaults into preferences.  See https://developer.android.com/guide/topics/ui/settings#Defaults
     PreferenceManager.setDefaultValues(this, R.xml.preferences, false);
 
+    // Registers this class as a listener for user preference changes.
+    PreferenceManager.getDefaultSharedPreferences(this).
```
cjhenck: Is there a risk of a memory leak here since you don't unregister? Is that automatic? I would guess it's OK if the non-service activity gets closed, but might be worth checking if you haven't already?

Android/app/src/main/java/app/intra/DnsVpnService.java
```diff
@@ -90,14 +91,30 @@ private void onResolved(Intent intent) {
       };
 
   @Override
-  public void onSharedPreferenceChanged(SharedPreferences preferences, String key) {
+  public synchronized void onSharedPreferenceChanged(SharedPreferences preferences, String key) {
```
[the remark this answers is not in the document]
bemasc: OK, I removed the synchronization.  It's technically not needed, because `restartVpn` always terminates the service.

Android/app/src/main/java/app/intra/ServerChooserFragment.java
bemasc: OK, done

Android/app/src/main/java/app/intra/ServerChooserFragment.java
bemasc: Done

Android/app/src/main/java/app/intra/ServerChooserFragment.java
bemasc: The problem is, entering a bare domain name is (1) very tempting and (2) not going to work (unless the server is actually running the API endpoint on `/`, which would be very strange).  Requiring a path at least makes this explicit, and hopefully reduces the frequency of user error.

Android/app/src/main/java/app/intra/MainActivity.java
bemasc: OK, I added an unregister call.

Android/app/src/main/java/app/intra/ServerChooserFragment.java
bemasc: In this design, I can't call `updateUI` here because `setValid` requires `Dialog` to exist, but it doesn't exist yet.

cjhenck: Or you could do a null check?  Anyway, your choice :)

Android/app/src/main/java/app/intra/ServerChooserFragment.java
cjhenck: Fair enough

Android/app/src/main/java/app/intra/ServerChooserFragment.java
bemasc: I switched this to a more MVC-ish design.  Using `onCreateView` as we discussed doesn't work, because the action buttons don't exist yet, but I found a later life cycle event that works.

PR #27, opened 2018-08-06
Show a warning when a VPN is already running

bemasc:
This change also disables reporting of "onRevoke" as a crash.
This "crash" has occurred thousands of times without any
corresponding user complaints, so it seems likely to be
occurring when the user activates a VPN that replaces Intra.
Hopefully, this text will help users understand this interaction
better.

### DIFF
--- a/Android/app/src/main/java/app/intra/MainActivity.java
+++ b/Android/app/src/main/java/app/intra/MainActivity.java
@@ -27,9 +27,12 @@ import android.graphics.PorterDuff;
 import android.net.ConnectivityManager;
 import android.net.LinkProperties;
 import android.net.Network;
+import android.net.NetworkCapabilities;
 import android.net.Uri;
 import android.net.VpnService;
 import android.os.Build;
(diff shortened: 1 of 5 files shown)


Android/app/src/main/java/app/intra/MainActivity.java
```diff
@@ -463,6 +472,34 @@ private String getSystemDnsServer() {
...
+    if (VERSION.SDK_INT >= VERSION_CODES.M) {
+      Network activeNetwork = connectivityManager.getActiveNetwork();
+      if (activeNetwork == null) {
+        return false;
+      }
+      NetworkCapabilities capabilities = connectivityManager.getNetworkCapabilities(activeNetwork);
+      return capabilities.hasTransport(NetworkCapabilities.TRANSPORT_VPN);
+    }
+    // For pre-M versions, return true if there's any network whose name looks like a VPN.
+    try {
```
cjhenck: Yikes!

PR #34, opened 2018-08-14
Remove all dynamic allocations from onDraw

bemasc:
This is strongly recommended by the Android IDE and docs,
to avoid triggering a garbage collection pause while
drawing.

### DIFF
--- a/Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
+++ b/Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
@@ -50,94 +50,81 @@ public class DnsQueryTracker {
     sync();
   }
 
-  public long getNumRequests() {
-    synchronized (this) {
-      return numRequests;
-    }
+  public synchronized long getNumRequests() {
+    return numRequests;
   }
 
-  public Queue<DnsTransaction> getRecentTransactions() {
-    synchronized (this) {
-      return new LinkedList<>(recentTransactions);
-    }
+  public synchronized Queue<DnsTransaction> getRecentTransactions() {
+    return new LinkedList<>(recentTransactions);
   }
 
   public synchronized void readInto(DnsActivityReader reader) {
     reader.read(Collections.unmodifiableCollection(recentActivity));
   }
 
-  public int countQueriesSince(long startTime) {
-    synchronized (this) {
-      // Linearly scan recent activity.  Due to the small scale (N ~ 300), a more efficient algorithm
-      // does not appear to be necessary.
-      int queries = recentActivity.size();
-      for (long time : recentActivity) {
-        if (time < startTime) {
-          --queries;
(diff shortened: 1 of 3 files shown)


Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
```diff
@@ -60,10 +62,8 @@ public long getNumRequests() {
     }
   }
 
-  public Queue<Long> getRecentActivity() {
-    synchronized (this) {
-      return new LinkedList<>(recentActivity);
-    }
+  public synchronized void readInto(DnsActivityReader reader) {
```
cjhenck: it seems like a bad idea to be using two different locking mechanisms here.  E.g. this isn't synchronizing on `this` while other methods are, and so recentActivity could be modified while this is running.

I'm fairly certain there's no risk of locking though.

Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
```diff
@@ -124,7 +124,7 @@ public void recordTransaction(DnsTransaction transaction) {
     }
   }
 
-  public void sync() {
+  public synchronized  void sync() {
```
cjhenck: Likewise, not sure why you need to have two different methods of synchronizing?

cjhenck:
UI code seems reasonable (to the extent that it doesn't seem to do anything weird) but I had some questions about the thread-safety aspects of the new model

Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
bemasc: Marking a method as `synchronized` is equivalent to locking `this`, so it's not two different mechanisms.

However, since it's evidently confusing, I've done a cleanup pass and converted all the methods.

Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
bemasc: Oops.  I started doing this cleanup before and then decided I couldn't justify it in a UI performance PR, but forgot to revert this bit.  Luckily, now that the reviewer has complained it's totally justified.

Android/app/src/main/java/app/intra/util/DnsQueryTracker.java
cjhenck: Ah, I looked at the java documentation to try to find out if that was the case and didn't see it anywhere.

cjhenck:
[Found the intrinsic lock documentation](https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html):

> When a thread invokes a synchronized method, it automatically acquires the intrinsic lock for that method's object and releases it when the method returns.

PR #36, opened 2018-08-15
Improve display of *.co.uk and similar eTLD names

bemasc:
This required adding a dependency on Guava, to use its API that
provides Public Suffix List functionality.


Android/app/src/main/java/app/intra/RecyclerAdapter.java
```diff
@@ -197,26 +198,9 @@ private void setExpanded(boolean expanded) {
...
-        }
-        last = labels[i];
-      }
-      if (i == 0) {
-        return last;
-      }
-      return labels[i - 1] + "." + last;
+    // Convert an FQDN like "www.example.co.uk." to an eTLD + 1 like "example.co.uk".
+    private String getETldPlus1(String fqdn) {
+      return InternetDomainName.from(fqdn).topPrivateDomain().toString();
```
cjhenck: Yay for libraries!

PR #50, opened 2018-08-29
State machine revamp

### LINKED ISSUE #49: Improve connection state machine
Currently, the VPN establishment process is somewhat messy, not well localized within the code, and does not do a good job of indicating its state in the UI.  A better approach would factor the state into the following components:
 * VPN: off, on
 * Controller: want vpn off, want vpn on
 * Connection: (none), new, last query succeeded, last query failed

This state then maps into what we display in the UI:

  VPN | Controller | Connection | *Button* | *Status* | *Extra text*
 -------- | --------------- | ---------------- | ------------ | --- | -------
off | want vpn off | (none) | Enable Intra | Exposed | 
off | want vpn on | (none) | Starting... | Exposed |
on | want vpn on | (none) | Disable Intra | Protected |
on | want vpn on | new | Disable Intra | Protected | Connecting...
on | want vpn on | last query succeeded | Disable Intra | Protected | Connected
on | want vpn on | last query failed | Disable Intra | Protected |  Warning: Queries are failing.  You may want to restart Intra or try a different server.
~~on~~ | ~~want vpn off~~ | * | ~~Stopping...~~ | ~~Protected~~ |

EDIT: Stopping is synchronous so there's no intermediate state there.

bemasc:
This moves connection state tracking to a model-view-controller
arrangement, and exposes additional useful information to the
user, such as whether queries have recently been failing.  This
is a prerequisite for proposed UI changes that would cause all
users to connect to the default server, because some users may
not be able to reach that server, and we currently have no clear
way to indicate the problem.

Fixes #49


Android/app/src/main/java/app/intra/DnsVpnController.java
```diff
@@ -0,0 +1,114 @@
...
+// Singleton class to maintain state related to VPN Tunnel service.
+public class DnsVpnController {
+
+  private static DnsVpnController dnsVpnServiceState;
+
+  public Object clone() throws CloneNotSupportedException {
+    throw new CloneNotSupportedException();
+  }
+
+  public static synchronized DnsVpnController getInstance() {
```
alalamav: Would it be cleaner to pass the context here instead of passing it in all methods?

Android/app/src/main/java/app/intra/DnsVpnState.java
```diff
@@ -0,0 +1,20 @@
+package app.intra;
+
+public final class DnsVpnState {
+  // Whether the user has requested that the VPN be active.  This is persistent state, sync'd to
+  // disk.
+  public final boolean requested;
```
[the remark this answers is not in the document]
bemasc: Done

Android/app/src/main/java/app/intra/DnsVpnController.java
bemasc: I tried this, but I got a compiler warning about holding a static reference to `context`.  As I understand it, the problem is that the statically held `context` is actually `MainActivity`, and we want to make sure that there are no long-lived references to `MainActivity`, so that it can be garbage-collected even though the service (and the `DnsVpnController` singleton) remain.

PR #53, opened 2018-08-30
Improve state transitions

bemasc:
This change bundles four improvements to transitions between states:

1. Restarting the VPN is now seamless, whereas previously it
   caused the VPN to pass through the unprotected state.
2. Queries that were canceled are no longer counted as failing.
3. Resetting a connection now performs a deeper reset.  This appears
   to be more reliable and should not be any slower.
4. DnsVpnController.start() is now idempotent, to ensure that we
   avoid double-starts.

### DIFF
--- a/Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
+++ b/Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
@@ -252,6 +252,10 @@ public class DnsResolverUdpToHttps extends Thread {
 
   private void sendResult(DnsTransaction transaction) {
     vpnService.recordTransaction(transaction);
+    // Update the connection state.  If the transaction succeeded, then the connection is working.
+    // If the transaction failed, then the connection is not working.
+    // If the transaction was canceled, then we don't have any new information about the status
+    // of the connection, so we don't send an update.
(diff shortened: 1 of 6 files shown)


Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
```diff
@@ -255,7 +255,7 @@ private void sendResult(DnsTransaction transaction) {
     DnsVpnController controller = DnsVpnController.getInstance();
     if (transaction.status == DnsTransaction.Status.COMPLETE) {
       controller.onConnectionStateChanged(vpnService, ServerConnection.State.WORKING);
-    } else {
+    } else if (transaction.status != DnsTransaction.Status.CANCELED) {
```
alalamav: Do we handle the other states?

bemasc: I added a comment to make this clearer
```
    // Update the connection state.  If the transaction succeeded, then the connection is working.
    // If the transaction failed, then the connection is not working.
    // If the transaction was canceled, then we don't have any new information about the status
    // of the connection, so we don't send an update.
```

PR #56, opened 2018-09-05
Implement a UDP-only DOH forwarder

bemasc:
This change does not yet make use of the forwarder, but it
includes tests to ensure that the forwarder is working.

### DIFF
--- a/Android/app/src/main/java/app/intra/util/LogWrapper.java
+++ b/Android/app/src/main/java/app/intra/util/LogWrapper.java
@@ -1,0 +1,24 @@
+package app.intra.util;
+
+import com.google.firebase.crash.FirebaseCrash;
+
+/**
+ * Wrapper for Firebase Crash Reporting.  This allows unit testing of classes that contain logging.
+ */
+public class LogWrapper {
+  public static void report(Throwable t) {
+    try {
+      FirebaseCrash.report(t);
+    } catch (IllegalStateException e) {
+      // This only occurs during unit tests.
+    }
+  }
+
+  public static void logcat(int i, String s, String s1) {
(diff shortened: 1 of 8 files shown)


Android/app/src/main/java/app/intra/util/LogWrapper.java
```diff
@@ -0,0 +1,24 @@
+package app.intra.util;
+
+import com.google.firebase.crash.FirebaseCrash;
+
+/**
+ * Wrapper for Firebase Crash Reporting.  This allows unit testing of classes that contain logging.
+ */
+public class LogWrapper {
```
alalamav: As an alternative, we could add a handler to [Java's root logger](https://docs.oracle.com/javase/7/docs/api/java/util/logging/package-summary.html) that logs to Firebase. That way we can use the default logger and initialize it with a fake in tests.

bemasc: This sounds interesting, but I'm not familiar with it.  Does it provide an easy way to segregate info that we only want to log locally?  There are some log lines that we definitely don't want to upload by accident.

alalamav: Yes, you can switch the destination based on the log level. We do a similar thing in [Outline](https://github.com/Jigsaw-Code/outline-client/blob/master/cordova-plugin-outline/android/java/org/outline/log/OutlineLogger.java#L43) to automatically log to Sentry.

bemasc: OK, interesting.  I will open an issue to track that improvement.

PR #58, opened 2018-09-11
Use tun2socks and sockslib to implement Intra

### LINKED ISSUE #19: Some downloads don't work when Intra is active

### LINKED ISSUE #51: Fix Chrome compatibility
Chrome appears to have activated an "async resolver" feature that [gets dns servers](https://chromium.googlesource.com/chromium/src/net/+/master/android/java/src/org/chromium/net/AndroidNetworkLibrary.java) directly from the OS using `getActiveNetwork()`.  Unfortunately, Intra's VPN is never the active network due to handling of the INTERNET capability, so Chrome gets the physical network's resolver IPs instead.  Chrome sends queries as UDP writes directly to those addresses, bypassing Intra.

To fix this, we can add the active network's DNS resolvers to our VPN's aperture.

bemasc:
On Android M (API 23) and later, this version uses a full-scope
VPN to intercept all traffic, instead of a split-horizon VPN that
only covers the DNS server.  The VPN re-emits all traffic unmodified
except for DNS, which is routed through DOH.

This includes a subset of the sockslib source code, binaries for tun2socks (taken
unmodified from Outline), and a suite of new classes to implement the
new VPN functionality.

Fixes #51
Fixes #19


Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
```diff
@@ -73,7 +79,8 @@ static void processQuery(ServerConnection serverConnection, DnsUdpQuery query,
      * @param responseWriter Receives the response
      */
     DnsResponseCallback(ServerConnection serverConnection, DnsUdpQuery request,
-                        DnsResponseWriter responseWriter) {
+                               DnsResponseWriter
```
alalamav: nit: formatting

Android/app/src/main/java/app/intra/DnsVpnAdapter.java
```diff
@@ -60,30 +69,140 @@
...
   DnsVpnService vpnService;
-  private final ParcelFileDescriptor tunFd;
+  @NonNull private final ParcelFileDescriptor tunFd;
   private final FileInputStream in;
   private final FileOutputStream out;
 
-  // Owners can safely mutate serverConnection at any time, including setting it to null.
-  @Nullable
-  ServerConnection serverConnection = null;
+  static DnsVpnAdapter get(DnsVpnService vpnService) {
```
alalamav: Could you rename this method to indicate that it establishes the VPN?

Android/app/src/main/java/app/intra/DnsVpnAdapter.java
```diff
@@ -207,35 +327,46 @@ private String getProtocolErrorMessage(byte protocol) {
...
-          dnsUdpQuery.destAddress,
-          dnsUdpQuery.sourceAddress,
-          rawUdpResponse);
-    } else {
-      ipPacket = new Ipv6Packet(
-          UDP_PROTOCOL,
-          dnsUdpQuery.destAddress,
-          dnsUdpQuery.sourceAddress,
-          rawUdpResponse);
+    if (transaction.response != null) {
```
alalamav: Early return may be cleaner.

Android/app/src/main/java/app/intra/GoogleServerDatabase.java
```diff
@@ -19,6 +19,8 @@
 import android.content.res.AssetManager;
 
 import android.util.Log;
+
```
alalamav: nit: extra space

Android/app/src/main/java/app/intra/LocalhostResolver.java
```diff
@@ -42,37 +44,37 @@
...
    * @param vpnService The VPN service that owns this server.
-   * @param serverConnection The connection to an upstream DOH server.
    * @return The localhost resolver, or null if it couldn't be started.
    */
-  public static LocalhostResolver get(DnsVpnService vpnService, ServerConnection serverConnection) {
+  public static LocalhostResolver get(DnsVpnService vpnService) {
     DatagramSocket socket = null;
     try {
-      SocketAddress bindaddr = new InetSocketAddress("localhost", 0);
+      InetAddress localhost = InetAddress.getByAddress(new byte[]{127, 0, 0, 1});
```
alalamav: Can you use `InetAddress.getLocalHost()`?

Android/app/src/main/java/app/intra/socks/SocksVpnAdapter.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+  }
+
+  // DNS resolver running on localhost.
+  private LocalhostResolver resolver;
+
+  // TUN device representing the VPN.
+  private ParcelFileDescriptor tunFd;
+
+  public static SocksVpnAdapter get(@NonNull DnsVpnService vpnService) {
```
alalamav: same suggestion as in DnsVpnAdapter

Android/app/src/main/java/app/intra/socks/UdpOverrideSocksHandler.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Override
+  public void doUDPAssociate(Session session, CommandMessage commandMessage) throws IOException {
+    // Construct a UDP relay server, just like upstream.
+    UDPRelayServer udpRelayServer =
+        new UDPRelayServer(((InetSocketAddress) session.getClientAddress()).getAddress(),
+            commandMessage.getPort());
+
+    // Set buffer size to 5 KB.  (The default is 5 MB, which causes OOM crashes.)
+    final int BUFFER_SIZE = 5 * 1024;
```
alalamav: Could you make these final variables static constants?

Android/app/src/main/java/app/intra/DnsVpnAdapter.java
```diff
@@ -60,30 +69,140 @@
 
   private static final int DNS_DEFAULT_PORT = 53;
 
+  // VPN parameters
+  private static final int VPN_INTERFACE_MTU = 32767;
```
alalamav: You could extract this variable to `VpnAdapter`, since it is also used by `SocksVpnAdapter`.

Android/app/src/main/java/app/intra/socks/SocksVpnAdapter.java
```diff
@@ -0,0 +1,177 @@
...
+    this.resolver = resolver;
+    this.tunFd = tunFd;
+  }
+
+  @Override
+  public void run() {
+    resolver.start();
+
+    // VPN parameters
+    final int DNS_PORT = 53;
```
alalamav: This constant is also used in `DnsVpnAdapter`.

Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
bemasc: Fixed.  Thanks!

Android/app/src/main/java/app/intra/DnsVpnAdapter.java
bemasc: We decided not to do this due to the final call to `recordTransaction`.

Android/app/src/main/java/app/intra/LocalhostResolver.java
bemasc: `getLocalHost()` seems scary.  The [docs](https://developer.android.com/reference/java/net/InetAddress#getLocalHost()) don't promise that it actually resolves to a "loopback" IP; they just say:

> Returns the address of the local host. This is achieved by retrieving the name of the host from the system, then resolving that name into an InetAddress.

[`getLoopbackAddress`](https://developer.android.com/reference/java/net/InetAddress#getLoopbackAddress()) actually does promise to do what we want, but on my phone it returns IPv6 localhost, which doesn't seem to work with tun2socks right now.

Android/app/src/main/java/app/intra/GoogleServerDatabase.java
bemasc: Fixed

Android/app/src/main/java/app/intra/DnsVpnAdapter.java
bemasc: Good idea.  Renamed from `get` to `establish`.

Android/app/src/main/java/app/intra/socks/SocksVpnAdapter.java
bemasc: Done

Android/app/src/main/java/app/intra/socks/UdpOverrideSocksHandler.java
bemasc: Done

Android/app/src/main/java/app/intra/DnsVpnAdapter.java
bemasc: Done

Android/app/src/main/java/app/intra/socks/SocksVpnAdapter.java
bemasc: Moved to `VpnAdapter`.

PR #71, opened 2018-09-20
New intro experience

bemasc:


Android/app/src/main/java/app/intra/IntroDialog.java
```diff
@@ -0,0 +1,187 @@
...
+import androidx.viewpager.widget.ViewPager;
+
+// Fragment representing the full-screen first-run experience.  This intro is launched by the main
+// activity.
+public class IntroDialog extends DialogFragment {
+  private static final int NUM_PAGES = 3;
+
+  // Avoid duplicate instances of this dialog.  This can occur when rotating the app, which would
+  // otherwise cause the main activity to launch an additional instance of this dialog.
+  private static boolean shown = false;
```
alalamav: nit: s/shown/isShown

Android/app/src/main/java/app/intra/IntroDialog.java
```diff
@@ -0,0 +1,187 @@
...
+      }
+    });
+
+    pager.addOnPageChangeListener(new TabListener(backButton, nextButton, acceptButton));
+
+    shown = true;
+    return welcomeView;
+  }
+
+  public static class Page0 extends Fragment {
```
alalamav: These class names and duplication seem like a smell to me.
Could we have a single class for the three pages? We could expose a property to set the one different string and in the case of `Page2`, a method to set the `details`?

Android/app/src/main/java/app/intra/IntroDialog.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Override
+    public int getCount() {
+      return pages.length;
+    }
+
+    @Override
+    public Fragment getItem(int position) {
+      return pages[position];
```
alalamav: Maybe check that the index is in bounds?

Android/app/src/main/java/app/intra/IntroDialog.java
bemasc: Done

Android/app/src/main/java/app/intra/IntroDialog.java
bemasc: OK, I've unified the pages into a single layout and fragment class.  PTAL.

Android/app/src/main/java/app/intra/IntroDialog.java
bemasc: Done.

alalamav:
Nice, LGTM!

PR #74, opened 2018-09-21
Fix RTL language support

bemasc:
In addition to cosmetic fixes, this also fixes a connection
issue due when connecting to dns.google.com in arabic-numeral
locales.

### DIFF
--- a/Android/app/src/main/java/app/intra/IntroDialog.java
+++ b/Android/app/src/main/java/app/intra/IntroDialog.java
@@ -17,7 +17,6 @@ package app.intra;
 
 import android.app.Activity;
 import android.graphics.drawable.Drawable;
-import android.os.Build;
 import android.os.Bundle;
 import android.text.method.LinkMovementMethod;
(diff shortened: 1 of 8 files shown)


Android/app/src/main/java/app/intra/IntroDialog.java
```diff
@@ -58,26 +58,38 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle
       savedInstanceState) {
     View welcomeView = inflater.inflate(R.layout.intro_pager, container);
 
+    // Right-to-left locale support.
+    boolean rtl = false;
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
```
alalamav: Is it worth pulling this logic to a util class?

bemasc: Done.  PTAL.

alalamav:
LGTM!

Issue #64, opened 2018-09-13
Restore IPv6 support


bemasc:
The new VPN core doesn't support IPv6.  We should fix that.

bemasc:
This had to be reverted due to an apparent double-free or other memory cleanup issue in tun2socks.

bemasc:
I had to revert this in #155.

PR #79, opened 2018-09-24
Update to the latest tun2socks

bemasc:
This change pulls the binary directly from Outline, to
benefit from its well-tested build environment.

This version of tun2socks contains IPv6 support, but we
can't enable it yet due to #64.


anzetal:
Hi. Does this mean if I use google DNS intra app, I will not be able to access IPv6 websites even if I have IPv6 connectivity

PR #85, opened 2018-09-26
Fix a crash when receiving an empty DNS response

bemasc:
Currently, if the server sends a successful HTTP response
with an empty body, Intra will try to rewrite the first two
bytes (which don't exist), resulting in a BufferOverflowException.

This change also includes a small refactor for clarity.

### DIFF
--- a/Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
+++ b/Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
@@ -22,6 +22,7 @@ import android.util.Log;
 import java.io.IOException;
 import java.net.Inet4Address;
 import java.net.SocketTimeoutException;
+import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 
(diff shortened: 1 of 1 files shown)


Android/app/src/main/java/app/intra/DnsResolverUdpToHttps.java
```diff
@@ -87,9 +88,14 @@ static void processQuery(ServerConnection serverConnection, DnsUdpQuery query,
     }
 
     // Writes |dnsRequestId| to |dnsResponse|'s ID header.
-    private void writeRequestIdToDnsResponse(byte[] dnsResponse, short dnsRequestId) {
+    private boolean writeRequestIdToDnsResponse(byte[] dnsResponse, short dnsRequestId) {
```
alalamav: Do you see any advantages in returning a boolean instead of throwing the exception?

bemasc: That seems like a good approach.  Done.

PR #86, opened 2018-09-27
Refactor SettingsFragment state handling

bemasc:
This is an attempt to resolve a null pointer exception that
was observed with the previous approach, by moving closer to
the standard AsyncTask design pattern.

### DIFF
--- a/Android/app/src/androidTest/java/app/intra/util/AppInfoTest.java
+++ b/Android/app/src/androidTest/java/app/intra/util/AppInfoTest.java
@@ -1,0 +1,57 @@
+package app.intra.util;
+
+import android.os.Parcel;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class AppInfoTest {
+  final String LABEL1 = "label1", LABEL2 = "label2";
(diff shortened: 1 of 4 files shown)


Android/app/src/androidTest/java/app/intra/util/AppInfoTest.java
```diff
@@ -0,0 +1,57 @@
...
+    assertEquals(1, NULL1.compareTo(APP1));
+    assertEquals(-1, APP2.compareTo(NULL2));
+    assertEquals(1, NULL2.compareTo(APP2));
+    assertEquals(-1, APP2.compareTo(NULL1));
+    assertEquals(1, NULL1.compareTo(APP2));
+    assertEquals(-1, NULL1.compareTo(NULL2));
+    assertEquals(1, NULL2.compareTo(NULL1));
+  }
+
+  private void roundtrip(AppInfo info) {
```
alalamav: This test method could benefit from a more descriptive name.

bemasc: Done.

PR #111, opened 2018-10-09
Don't block the UI thread during bootstrap

### LINKED ISSUE #103: Intra can become non-responsive if the default server isn't reachable
Observed behavior:
If the default server isn't reachable during the initial setup flow, the OS can interpret intra as non-responsive.  Users also can't cancel the connection and are not offered another choice.

Expected behavior:
1. Remain responsive, and allow users to cancel a connection and change the desired server.
2. Time-out more quickly with a detailed notification to suggest changing the server.

bemasc:
Fixes #103

### DIFF
--- a/Android/app/src/main/java/app/intra/DnsVpnService.java
+++ b/Android/app/src/main/java/app/intra/DnsVpnService.java
@@ -53,12 +53,34 @@ public class DnsVpnService extends VpnService implements NetworkManager.NetworkL
   private static final String LOG_TAG = "DnsVpnService";
   private static final int SERVICE_ID = 1; // Only has to be unique within this app.
   private static final String CHANNEL_ID = "vpn";
+  private static final String NO_PENDING_CONNECTION = "This value is not a possible URL.";
 
+  // The network manager is populated in onStartCommand.  Its main function is to enable delayed
+  // initialization if the network is initially disconnected.
   private NetworkManager networkManager;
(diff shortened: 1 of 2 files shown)


Android/app/src/main/java/app/intra/DnsVpnService.java
```diff
@@ -53,13 +53,35 @@
   private static final String LOG_TAG = "DnsVpnService";
   private static final int SERVICE_ID = 1; // Only has to be unique within this app.
   private static final String CHANNEL_ID = "vpn";
+  private static final String NO_PENDING_CONNECTION = "This value is not a possible URL.";
```
alalamav: nit: maybe use a more standard sentinel value like `""` or `null`?

Android/app/src/main/java/app/intra/DnsVpnService.java
```diff
@@ -53,13 +53,35 @@
...
 
+  // The network manager is populated in onStartCommand.  Its main function is to enable delayed
```
alalamav: nice commenting ;)

Android/app/src/main/java/app/intra/DnsVpnService.java
bemasc: Unfortunately that doesn't work here because `null` and `""` are both treated as meaning "default server".

PR #123, opened 2018-10-16
Count download bytes and early reset events

bemasc:
This should make it easier to monitor and improve the reliability
of the network subsystem.

### DIFF
--- a/Android/app/src/main/java/app/intra/socks/ReliabilityMonitor.java
+++ b/Android/app/src/main/java/app/intra/socks/ReliabilityMonitor.java
@@ -1,0 +1,117 @@
+package app.intra.socks;
+
+import android.content.Context;
+import android.os.Bundle;
+import android.os.SystemClock;
+import app.intra.util.Names;
+import com.google.firebase.analytics.FirebaseAnalytics;
+import com.google.firebase.analytics.FirebaseAnalytics.Param;
+import java.net.Socket;
+import sockslib.server.io.Pipe;
+import sockslib.server.io.PipeListener;
+import sockslib.server.io.SocketPipe;
+import sockslib.server.listener.PipeInitializer;
+
+/**
+ * This is a pipe initializer that monitors post-connection socket reliability.
+ */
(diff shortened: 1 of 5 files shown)


Android/app/src/main/java/app/intra/socks/ReliabilityMonitor.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private static int getPort(Pipe pipe) {
+      final Socket remoteSocket;
+      if (SocketPipe.INPUT_PIPE_NAME.equals(pipe.getName())) {
+        remoteSocket = (Socket) pipe.getAttribute(SocketPipe.ATTR_SOURCE_SOCKET);
+      } else if (SocketPipe.OUTPUT_PIPE_NAME.equals(pipe.getName())){
+        remoteSocket = (Socket) pipe.getAttribute(SocketPipe.ATTR_DESTINATION_SOCKET);
+      } else {
+        remoteSocket = null;
```
alalamav: You could return here, though I see the reason too keep the explicit check.

Android/app/src/main/java/app/intra/socks/ReliabilityMonitor.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public Pipe initialize(Pipe pipe) {
+    pipe.addPipeListener(new ReliabilityListener(context));
+    return pipe;
+  }
+
+  /**
+   * This class converts pipe lifecycle callbacks into network reliability monitoring events.
```
alalamav: Could you add a comment about the event (data structure) being sent?

Android/app/src/main/java/app/intra/socks/ReliabilityMonitor.java
bemasc: Yeah, my thinking was that it would better to avoid two instances of `return -1`, i.e. the magic constant should only appear once.

PR #130, opened 2018-10-25
Make UI more consistent when DNS over TLS is on

### LINKED ISSUE #129: Correct "off" state dns protocol when Android Private DNS is enabled
Right now the protocol field says "dns" unconditionally, which could be confusing on phones that have DNS over TLS enabled.

bemasc:
Fixes #129


bemasc:
This has been merged into #131.

PR #131, opened 2018-10-25
Add info pages for each of the detailed info tiles

bemasc:
This makes all the info tiles clickable, revealing a detailed explanation of each tile.

Note that the strings have already been reviewed.


cjhenck:
LGTM

PR #143, opened 2018-12-07
Reorganize packages

bemasc:
This change groups files into more sensible Java packages.
Some files have been renamed, and many methods have been
made package-private.

Text changes were largely automated by IntelliJ.


bemasc:
@alalamav I recommend not looking at the contents of the files, which are essentially just automated changes and renames.  To review this CL, I would recommend just looking at the organization of files and seeing if anything should be in a different package, or if you can think of better names for any of the classes.

Android/app/src/main/java/app/intra/sys/DnsActivityReader.java
```diff
@@ -0,0 +1,31 @@
...
+package app.intra.sys;
+
+import java.util.Collection;
+
+/**
+ * Visitor interface for DnsQueryTracker's activity data.  This allows a class to get access
+ * to the query activity without making a copy, while maintaining DnsQueryTracker's
+ * synchronization.
+ */
+public interface DnsActivityReader {
```
alalamav: The "reader" part of the class is a bit confusing to me. What does it read? It seems to me that "track" or something similar could be clearer.
(Also, what do you mean by visitor interface?)

bemasc: OK, thanks for the comment.  I renamed this and the related components, and improved the comments.

PR #144, opened 2018-12-10
Move to a shorter, more adaptive timeout for retry

bemasc:
This should make the retry much more usable, while still
being unlikely to trigger on sockets that would succeed
without retry.


bemasc:
Thanks for the suggestions. Done!

PR #173, opened 2019-05-06
Make it easier to change servers when one breaks

bemasc:
Currently, it's not very clear what to do when queries
are failing.  This change makes it clearer that users
have the option to switch servers, which might help.
When queries are failing, a button appears that gives
users a shortcut to the server settings dialog.


Android/app/src/main/java/app/intra/ui/MainActivity.java
```diff
@@ -572,6 +576,10 @@ private void syncDnsStatus() {
     final TextView indicatorText = controlView.findViewById(R.id.indicator);
     indicatorText.setText(status.activationRequested ? R.string.indicator_on : R.string.indicator_off);
 
+    // Hide server change button by default
+    final Button changeServerButton = controlView.findViewById(R.id.change_server_button);
+    changeServerButton.setVisibility(View.INVISIBLE);
```
alalamav: `syncDnsStatus ` will get called when the new connection is set up, right? I was wondering if we needed to explicitly hide it.

bemasc: Yes.  If the `VpnState` changes, `syncDnsStatus` will get called again with the new state.

PR #184, opened 2019-05-22
Enable proguard for release builds

bemasc:
This is the recommended build configuration for Android.
It's also required because otherwise (thanks to Guava) we
have more than 65536 methods, which is the limit for one
Android DEX file, and support for multi-DEX APKs on older
Android versions is limited.  Proguard removes all the
unused methods, keeping us under the DEX file limit (and
also reducing the APK size).

### DIFF
--- a/Android/app/build.gradle
+++ b/Android/app/build.gradle
@@ -47,8 +47,10 @@ android {
     }
     buildTypes {
         release {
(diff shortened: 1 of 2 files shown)


bemasc:
Just tested on Android 4.1, and didn't see any errors like that.

PR #192, opened 2019-06-18
Report events when switching servers

bemasc:
This should help us understand whether users are
able to recover from server reachability failures.


Android/app/src/main/java/app/intra/ui/MainActivity.java
```diff
@@ -602,6 +603,7 @@ private void syncDnsStatus() {
         changeServerButton.setOnClickListener((View view) -> {
           chooseView(R.id.settings);
           openServerSetting();
+          FirebaseAnalytics.getInstance(this).logEvent(Names.SERVER_FAIL_BUTTON.name(), null);
```
alalamav: Would it be useful (privacy-compliant) to log the failed server name?

bemasc:
This conflicts with #195, so I'm going to withdraw this and we can reconsider what events make sense after that change is complete.

PR #198, opened 2019-07-12
Switch Google to RFC DoH by default

bemasc:
This change still falls back to pre-standard DoH if
RFC DoH isn't working.

This should future-proof Intra against the turndown of
Google's pre-standard DoH service.

### DIFF
--- a/Android/app/src/main/java/app/intra/net/doh/ServerConnectionFactory.java
+++ b/Android/app/src/main/java/app/intra/net/doh/ServerConnectionFactory.java
@@ -22,6 +22,7 @@ import android.util.Log;
 import app.intra.R;
 import app.intra.net.doh.google.GoogleServerConnection;
 import app.intra.net.doh.google.GoogleServerDatabase;
+import app.intra.sys.PersistentState;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


Android/app/src/main/java/app/intra/net/doh/ServerConnectionFactory.java
```diff
@@ -43,14 +44,10 @@ public ServerConnectionFactory(Context context) {
...
    */
-  public static boolean equalUrls(String url1, String url2) {
-    // null and "" are equivalent, representing the default server.
-    if (url1 == null) {
-      url1 = "";
-    }
-    if (url2 == null) {
-      url2 = "";
-    }
+  public boolean equalUrls(String url1, String url2) {
```
alalamav: I wonder if this method better remain static.

Android/app/src/main/java/app/intra/sys/IntraVpnService.java
```diff
@@ -60,6 +58,9 @@
   private static final String WARNING_CHANNEL_ID = "warning";
   private static final String NO_PENDING_CONNECTION = "This value is not a possible URL.";
 
+  // Factory for producing DoH connections.
+  private final ServerConnectionFactory dohFactory = new ServerConnectionFactory(this);
```
alalamav: Do you need this to be an instance variable? I only see it used in `updateServerConnection`.

bemasc: Done

Android/app/src/main/java/app/intra/net/doh/ServerConnectionFactory.java
bemasc: OK, done.

bemasc:
> Will Google continue to support pre-standard DOH indefinetely?

Yes, but based on [the migration schedule](https://developers.google.com/speed/public-dns/docs/doh/migration) I think we should probably plan to switch to pure RFC DOH after 2019-09-23.

bemasc:
After thinking some more about the migration sequence, and the current code's behavior, I think the net effect of this change is just to slow things down.  The current code should work at least until June of 2020, so I think it's probably better to wait until the September transition.  Then we can remove all the pre-standard DoH code and move to RFC DoH for Google.

alalamav:
Sounds good to me.

PR #201, opened 2019-07-15
Increment dependency versions

bemasc:
The biggest change here is the temporary addition of
androidx.fragment 1.1.0-rc1, to get a fix for a bug
that is causing crashes for some users.


alalamav:
You may want to look into CI failing to fetch a dependency.

bemasc:
Thanks.  I forgot to try the lint target, which turns out to need different dependencies (and also imposes a new linter requirement).  Fixed.

PR #211, opened 2019-08-16
Enable A/B testing of go-tun2socks

bemasc:


Android/app/src/main/java/app/intra/net/socks/GoIntraListener.java
```diff
@@ -0,0 +1,93 @@
...
+      // Prepare an EARLY_RESET event to collect metrics on success rates for splitting:
+      // - BYTES : Amount uploaded before reset
+      // - CHUNKS : Number of upload writes before reset
+      // - TIMEOUT : Whether the initial connection failed with a timeout.
+      // - SPLIT : Number of bytes included in the first retry segment
+      // - RETRY : 1 if retry succeeded, otherwise 0
+      Bundle resetEvent = new Bundle();
+      resetEvent.putInt(Names.BYTES.name(), retry.getBytes());
+      resetEvent.putInt(Names.CHUNKS.name(), retry.getChunks());
+      resetEvent.putInt(Names.TIMEOUT.name(), retry.getTimeout() ? 1 : 0);
```
alalamav: Is there an advantage in using `putInt` vs `putBoolean`?

Android/app/src/main/java/app/intra/net/socks/GoVpnAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Locale;
+import tun2socks.IntraTunnel;
+import tun2socks.Tun2socks;
+
+/**
+ * This is a VpnAdapter that captures all traffic and routes it through a go-tun2socks instance with
+ * custom logic for Intra.
+ */
+public class GoVpnAdapter extends VpnAdapter {
+  private static final String LOG_TAG = "SocksVpnAdapter";
```
alalamav: Maybe rename to `"GoVpnAdapter"` to distinguish it.

Android/app/src/main/java/app/intra/net/socks/GoVpnAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+
+    // VPN parameters
+    final String fakeDnsIp = LanIp.DNS.make(IPV4_TEMPLATE);
+    final String fakeDns = fakeDnsIp + ":" + DNS_DEFAULT_PORT;
+
+    Result r = TLSProbe.run(context);
+    LogWrapper.log(Log.INFO, LOG_TAG, "TLS probe result: " + r.name());
+    boolean alwaysSplitHttps = r == Result.TLS_FAILED;
+
+    String trueDns = resolver.getAddress().toString().substring(1);
```
alalamav: nit: why do you discard the first character? Does the address string start with a slash?

Android/app/src/main/java/app/intra/net/socks/GoIntraListener.java
bemasc: The immediate reason is that these attributes are already in use, so it's probably a bad idea to change their types now.  The original reason is that Firebase only knows about numeric and string-valued attributes, and places a limit on the number of string-valued attributes per project.

Android/app/src/main/java/app/intra/net/socks/GoVpnAdapter.java
bemasc: Yes.  I added a comment to explain the situation.

Android/app/src/main/java/app/intra/net/socks/GoVpnAdapter.java
bemasc: Thanks, you're right.  Done.

PR #213, opened 2019-08-19
Revert "Split APK by ABI"

bemasc:
The Play Store doesn't permit multiple APKs for different architectures to share a version code, so this splitting configuration is basically not usable in practice.  We'll just have to stick with the unified APK until we can get bundles working.


bemasc:
@alalamav FYI

PR #222, opened 2019-10-07
Clarify legacy config upgrade logic

bemasc:
This should have no effect on behavior.

### DIFF
--- a/Android/app/src/main/java/app/intra/sys/PersistentState.java
+++ b/Android/app/src/main/java/app/intra/sys/PersistentState.java
@@ -78,29 +78,30 @@ public class PersistentState {
 
     // There is no URL setting, so read the legacy server name.
     SharedPreferences settings = getInternalState(context);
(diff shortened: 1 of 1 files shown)


Android/app/src/main/java/app/intra/sys/PersistentState.java
```diff
@@ -78,29 +78,30 @@ public static void syncLegacyState(Context context) {
...
-      if (domain.equals(parsed.getHost())) {
-        url = urls[i];
-        break;
+    } else {
+      // In practice, we expect that domain will always be cloudflare-dns.com at this point, because
+      // that was the only other option in the relevant legacy versions of Intra.
+      // Look for the corresponding URL among the builtin servers.
+      for (String u : urls) {
+        Uri parsed = Uri.parse(u);
+
```
alalamav: nit: extra newline

bemasc: Removed

PR #226, opened 2019-11-01
Include mobile country code in Analytics events

bemasc:
This should reduce country reporting errors in Firebase.
This change also reorganizes event reporting to reduce
boilerplate.


Android/app/src/main/java/app/intra/sys/AnalyticsEvent.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  // Parameters sent in Analytics events
+  public enum Params {
+    BYTES,
+    CHUNKS,
+    DOWNLOAD,
+    DURATION,
+    LATENCY,
+    MCC,
```
fortuna: We need to be clearer about what MCC you are referring to: the connected network or the home network? Let's make that clear in the names.

Android/app/src/main/java/app/intra/sys/AnalyticsEvent.java
```diff
@@ -0,0 +1,120 @@
...
+  // across all the analytics events.
+  private static Integer mcc = null;
+
+  // Get the system's mobile country code, or 0 if it could not be determined.
+  // After the first call to this function, all subsequent calls will return immediately.
+  private int getMCC() {
+    if (mcc != null) {
+      return mcc;
+    }
+    int gsmMCC = context.getResources().getConfiguration().mcc;
```
fortuna: Is this the recommended way, versus using the Telephony api?

fortuna: It seems the logic for getting the MCC from a context is quite application-independent, so I recommend moving it out of the application code. I guess that would be its own class. Or at least make it a static method that makes it clear it doesn't need the AnalyticsEvent state.

Android/app/src/main/java/app/intra/sys/AnalyticsEvent.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  // Mobile Country Code, or 0 if it could not be determined.  This is used to memoize getMCC()
+  // across all the analytics events.
+  private static Integer mcc = null;
+
+  // Get the system's mobile country code, or 0 if it could not be determined.
+  // After the first call to this function, all subsequent calls will return immediately.
+  private int getMCC() {
+    if (mcc != null) {
```
fortuna: I suggest you pass the home MCC in the constructor instead, and provide a factory function that takes the context. It will make the class a lot simpler, more testable and more reusable.

fortuna: Is there any benefit of lazy initiation?

Android/app/src/main/java/app/intra/net/go/GoIntraListener.java
```diff
@@ -62,13 +62,13 @@ public void onTCPSocketClosed(TCPSocketSummary summary) {
...
 
-    Bundle bytesEvent = new Bundle();
-    bytesEvent.putLong(Names.UPLOAD.name(), summary.getUploadBytes());
-    bytesEvent.putLong(Names.DOWNLOAD.name(), summary.getDownloadBytes());
-    bytesEvent.putInt(Names.PORT.name(), summary.getServerPort());
-    bytesEvent.putInt(Names.TCP_HANDSHAKE_MS.name(), summary.getSynack());
-    bytesEvent.putInt(Names.DURATION.name(), summary.getDuration());
-    FirebaseAnalytics.getInstance(vpnService).logEvent(Names.BYTES.name(), bytesEvent);
+    new AnalyticsEvent(vpnService)
+        .put(Params.UPLOAD, summary.getUploadBytes())
```
fortuna: I'm wondering if it makes sense to create factory functions or static methods like SendBytesEvent(upload_bytes, download_bytes, port, tcpHandshakeMs, duration)

That would hide all the Params details from the application code and provide some type safety. Also, most of the time you set the params and send in the same call.

Even though it may result in a little more code, you'll be isolating complexity and the application code will be more readable and simpler (e.g. 1 line instead of 7 to send an event).

Android/app/src/main/java/app/intra/sys/AnalyticsEvent.java
bemasc: I changed the code to only report when they agree, to minimize confusion.

Android/app/src/main/java/app/intra/sys/AnalyticsEvent.java
bemasc: I switched to the Telephony API, in order to get both SIM and Network codes.

bemasc: OK, I've moved it into a new class.

Android/app/src/main/java/app/intra/sys/AnalyticsEvent.java
bemasc: I've moved the context into a factory method as you suggested.

Android/app/src/main/java/app/intra/net/go/GoIntraListener.java
bemasc: Done!

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
```diff
@@ -0,0 +1,75 @@
...
+import android.telephony.TelephonyManager;
+import java.lang.reflect.Method;
+
+/**
+ * Static utility function for getting the Mobile Country Code.
+ */
+class MobileCountryCode {
+  // Mobile Country Code, or 0 if it could not be determined.  This is used to memoize get()
+  // across all the analytics events.
+  private static Integer mcc = null;
```
fortuna: Let's move the memoization to the AnalyticsWrapper instead, since this is higher level logic that can cause confusion at a low level (really an application optimization)

The logic of only using the country if they match could go in the AnalyticsWrapper too, since that's clearly application logic.

Android/app/src/main/java/app/intra/sys/firebase/AnalyticsWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+import androidx.annotation.NonNull;
+import app.intra.sys.firebase.BundleBuilder.Params;
+import com.google.firebase.analytics.FirebaseAnalytics;
+
+/**
+ * Library of Analytics events reported by Intra.  In addition to improving the reporting API,
+ * this class also attaches the Mobile Country Code (MCC) to each event, which indicates the client
+ * country, to compensate for country classification errors in Firebase.
+ */
+public class AnalyticsWrapper {
```
fortuna: This became a nice documentation of all the events we log. Cool!

Android/app/src/main/java/app/intra/sys/firebase/AnalyticsWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+    TRY_ALL_DIALOG,
+    TRY_ALL_FAILED,
+    TRY_ALL_REQUESTED,
+    UDP,
+  }
+
+  // Mobile Country Code, or 0 if it is unknown.
+  private final int mcc;
+  private final FirebaseAnalytics analytics;
+  private AnalyticsWrapper(int mcc, FirebaseAnalytics analytics) {
```
fortuna: I think we should move to ISO country codes instead, since MCC is a pain to convert.

Android/app/src/main/java/app/intra/sys/firebase/AnalyticsWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+    this.analytics = analytics;
+  }
+
+  public static AnalyticsWrapper get(Context context) {
+    return new AnalyticsWrapper(MobileCountryCode.get(context),
+        FirebaseAnalytics.getInstance(context));
+  }
+
+  private void log(Events e, @NonNull BundleBuilder b) {
+    b.put(Params.MCC, mcc);
```
fortuna: We need a name here that conveys:

- It's an ISO country code, not an MCC
- It's the home operator and network operator, set when the user is connected to its home cell network

It would also be useful to better understand when those things are set. I see 3 states:
- user on cell network at home
- user on cell network while roaming
- user not on cell network (e.g. wifi)

What do you think about this:
- `CELL_HOME_COUNTRY`: the "sim operator country". Set if the phone has a SIM card or cdma configured
- `CELL_NETWORK_COUNTRY` (maybe `CELL_CONNECTED_COUNTRY`? `CELL_NET_COUNTRY`?): Empty if not available (not on a cell network), set to the country if available and == `CELL_HOME_COUNTRY` or set to `ZZ` if it differs from `CELL_HOME_COUNTRY`, for privacy

This way we'll be able to better understand when those things are set, and the meaning of the parameters will be clearer.

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
```diff
@@ -0,0 +1,75 @@
...
+    if (mcc != null) {
+      return mcc;
+    }
+    TelephonyManager telephonyManager =
+        (TelephonyManager)context.getSystemService(Context.TELEPHONY_SERVICE);
+    Integer simMCC = extractMCC(telephonyManager.getSimOperator());
+    if (simMCC != null) {
+      // Require the SIM and network country codes to match before reporting a country, to avoid
+      // confusion when they disagree (e.g. international roaming).
+      Integer networkMCC = extractMCC(telephonyManager.getNetworkOperator());
```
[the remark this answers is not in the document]
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
```diff
@@ -0,0 +1,75 @@
...
+      }
+      return mcc;
+    }
+    // User appears to be non-GSM.  Try CDMA.
+    try {
+      // Get the system properties by reflection.
+      Class<?> systemPropertiesClass = Class.forName("android.os.SystemProperties");
+      Method get = systemPropertiesClass.getMethod("get", String.class);
+      String cdmaOperator = (String)get.invoke(systemPropertiesClass,
+          "ro.cdma.home.operator.numeric");
```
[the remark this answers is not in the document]
bemasc: Thanks.  I think that doesn't affect this code.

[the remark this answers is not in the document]
bemasc: That constant is also only accessible by reflection, and is not guaranteed to be stable, so I think we're better off using the setting's name.

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
```diff
@@ -0,0 +1,75 @@
...
+   * Gets the system's mobile country code, or 0 if it could not be determined.
+   * After the first call to this function, all subsequent calls will return immediately.
+   */
+  static int get(Context context) {
+    if (mcc != null) {
+      return mcc;
+    }
+    TelephonyManager telephonyManager =
+        (TelephonyManager)context.getSystemService(Context.TELEPHONY_SERVICE);
+    Integer simMCC = extractMCC(telephonyManager.getSimOperator());
```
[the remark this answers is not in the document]
bemasc: It looks like that was added in ~2016, so it should work for newer devices.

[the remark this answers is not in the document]
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Static utility function for getting the Mobile Country Code.
+ */
+class MobileCountryCode {
+  // Mobile Country Code, or 0 if it could not be determined.  This is used to memoize get()
+  // across all the analytics events.
+  private static Integer mcc = null;
+
+  // Extracts the MCC from an MCC+MNC string.
+  private static Integer extractMCC(String operator) {
```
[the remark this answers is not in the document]
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
```diff
@@ -0,0 +1,75 @@
...
+      return Integer.valueOf(operator.substring(0, 3));
+    }
+    return null;
+  }
+
+  /**
+   * Gets the system's mobile country code, or 0 if it could not be determined.
+   * After the first call to this function, all subsequent calls will return immediately.
+   */
+  static int get(Context context) {
```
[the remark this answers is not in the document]
bemasc: Done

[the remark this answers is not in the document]
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/MobileCountryCode.java
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/AnalyticsWrapper.java
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/AnalyticsWrapper.java
bemasc: I went with `COUNTRY` (the consensus country, or empty string if there is none) and `NETWORK` (the network type).  I think that's the minimal solution that gives us the necessary information and lets us exclude outlier events.

Android/app/src/main/java/app/intra/sys/firebase/CountryCode.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @NonNull String getNetworkCountry() {
+    return networkCountry;
+  }
+
+  /**
+   * @return The consensus country code, or "" if there isn't one.
+   */
+  @NonNull String getCombinedCountry() {
```
fortuna: I'm ok with providing a combined country, provided we know which country we are using. 
We need to better understand the data before trusting it. We need to see how those country parameters perform in the wild.

An alternative to my previous proposal is to add a new parameter that tells which country you picked. For example: COUNTRY_DEVICE, COUNTRY_NETWORK, COUNTRY_DEVICE_AND_NETWORK.

Though I have the impression the combined country will make it harder to pick different combinations. I suggested reporting the two locations instead (with the home or network country anonymized in case of roaming), because it's easy to apply filters like device_country == network_country. Or ip_country == network_country.

bemasc: OK, done.

Issue #154, opened 2019-01-19
Quick settings tile


maciejby:
Faster way to activate Intra

PR #233, opened 2019-11-14
Add quick settings tile. 

### LINKED ISSUE #194: Add a toggle in Quick Settings for faster access
It would be way more convenient to switch Intra on or off if there was a toggle in Quick Settings for it. Just like what CloudFlare DNS app has implemented:
Screenshot_20190621-103410

AryanNajafi:
Fixes #194 #154


bemasc:
:+1: Thanks a lot!

PR #235, opened 2019-11-27
Expand null URL before passing it to Go

bemasc:
This was causing total failure when using Go-DOH in fresh installs
and installs that had never adjusted the server setting.

### DIFF
--- a/Android/app/src/main/java/app/intra/net/go/GoVpnAdapter.java
+++ b/Android/app/src/main/java/app/intra/net/go/GoVpnAdapter.java
@@ -20,10 +20,13 @@ import android.os.Build;
 import android.os.ParcelFileDescriptor;
 import android.util.Log;
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import app.intra.net.VpnAdapter;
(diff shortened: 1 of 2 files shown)


bemasc:
Sorry, I made some additional changes here after you reviewed.  Could you take one more look?

Android/app/src/main/java/app/intra/net/go/GoVpnAdapter.java
```diff
@@ -106,10 +109,22 @@ public synchronized void start() {
...
+    if (RemoteConfig.getUseGoDoh()) {
+      try {
+        transport = makeDohTransport(dohURL);
+      } catch (Exception e) {
+        // Fall back to ServerConnection if Go-DOH setup fails.
+        // TODO: Report this failure properly instead of silently falling back.
+        LogWrapper.logException(e);
+        transport = null;
+      }
+    } else {
```
alalamav: `transport` is already `null` if `!RemoteConfig.useGoDoh()`.

bemasc: OK, I made this assignment more compact.

PR #239, opened 2019-12-06
Permit full-VPN tunnel mode on Android pre-M (23)

bemasc:
This change does not affect Intra's default behavior.  It adds a
remote configuration option that can be used to test full-VPN support
on versions below API 23.  In order to support versions below API
21, this change adds support for socket protection.

If this configuration works well, we should be able to remove OkHttp
and related codepaths.


Android/app/src/main/java/app/intra/sys/NetworkManager.java
```diff
@@ -113,4 +121,60 @@ private static boolean isConnectedNetwork(NetworkInfo networkInfo) {
...
+    };
+
+    List<InetAddress> resolvers = new ArrayList<>();
+    if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
+      LogWrapper.log(Log.ASSERT, LOG_TAG, "This function should never be called in L+.");
+      return resolvers;
+    }
+
+    try {
+      // LinkProperties ConnectivityManager.getLinkProperties(int type) (removed in Lollipop)
```
alalamav: Is there a way to target pre-lollipop APIs (i.e. through annotations) to avoid using reflection?

Android/app/src/main/java/app/intra/sys/firebase/RemoteConfig.java
```diff
@@ -36,16 +38,50 @@
...
+      // InetAddress.getAllByName(), which cannot be protected.)
+      return true;
+    }
+
     try {
-      return !FirebaseRemoteConfig.getInstance()
-          .getBoolean("disable_go_doh");
+      // go_min_version is the minimum Android version that should use the GoVpnAdapter (i.e. the
+      // full tunnel configuration).
+      long minVersion = FirebaseRemoteConfig.getInstance().getLong("go_min_version");
```
alalamav: Should the min Go version be defined in source instead than remotely?

Android/app/src/main/java/app/intra/sys/NetworkManager.java
bemasc: These APIs weren't public pre-Lollipop.  Well, to be precise, the `getLinkProperties` method and the entire `LinkProperties` class were `public`, but annotated with [`@hide`](https://stackoverflow.com/questions/17035271/what-does-hide-mean-in-the-android-source-code), which [excludes them](https://commonsware.com/blog/2018/01/18/think-hard-about-hide.html) from Javadoc and `android.jar`, which is what the compiler is compiling against.  They weren't part of the "Android API".

I'm not aware of a way to access them other than reflection.  In principle, I suppose one could supplement `android.jar` with another "platform jar" exposing these additional APIs, but I've never heard of such an approach being used.

Android/app/src/main/java/app/intra/sys/firebase/RemoteConfig.java
bemasc: The goal of remote configuration is to enable A/B testing and remote-config experiments, to test whether this codepath really works well in practice.  In principle (and in the available emulators), it works on all versions, but I have some concerns about memory usage and unexpected system behaviors.

Once we've established a min version (hopefully 15, which is Intra's min version overall), we can remove this code (and also the OkHttp DoH client).

Android/app/src/main/java/app/intra/sys/NetworkManager.java
alalamav: I see. Makes sense to use reflection.

Android/app/src/main/java/app/intra/sys/firebase/RemoteConfig.java
alalamav: Got it!

PR #243, opened 2020-01-13
Update dependency versions

bemasc:
The only changes required beyond version updates are:
 * A layout tweak to prevent extra scrollbars from appearing
 * A change to the proguard rules


bemasc:
I've been dogfooding this for 24 hours.  It seems to be working fine.

PR #255, opened 2020-03-26
Major cleanup.  Remove OkHttp and split mode.

bemasc:
This also removes the remote configuration setup code, since there are no longer any remotely configurable settings in this version.

This change is designed as nearly a straight deletion for ease of review.  It will need to be followed by a refactor to clean up some layering that is no longer required.


Android/app/src/main/java/app/intra/net/go/GoVpnAdapter.java
```diff
@@ -112,20 +106,19 @@ private void connectTunnel() {
...
-    String trueDns = resolver.getAddress().toString().substring(1);
     listener = new GoIntraListener(vpnService);
     String dohURL = PersistentState.getServerUrl(vpnService);
 
     try {
       LogWrapper.log(Log.INFO, LOG_TAG, "Starting go-tun2socks");
-      Transport transport = RemoteConfig.getUseGoDoh() ? makeDohTransport(dohURL) : null;
-      final IntraTunnel t = Tun2socks.connectIntraTunnel(tunFd.getFd(), fakeDns, trueDns, trueDns,
+      Transport transport = makeDohTransport(dohURL);
+      final IntraTunnel t = Tun2socks.connectIntraTunnel(tunFd.getFd(), fakeDns,
```
alalamav: nit: you could assign `tunnel` directly.

Android/app/src/main/java/app/intra/sys/firebase/RemoteConfig.java
```diff
@@ -1,90 +0,0 @@
...
-import com.google.android.gms.tasks.Task;
-import com.google.android.gms.tasks.Tasks;
-import com.google.firebase.remoteconfig.FirebaseRemoteConfig;
-
-/**
- * Utility class for initializing Firebase Remote Config.  Remote Configuration allows us to conduct
- * A/B tests of experimental functionality, and to enable or disable features without having to
- * release a new version of the app.
- */
-public class RemoteConfig {
```
alalamav: Would it be useful to retain a generic version of this class for future experiments?

Android/app/src/main/java/app/intra/net/go/GoVpnAdapter.java
bemasc: Done

Android/app/src/main/java/app/intra/sys/firebase/RemoteConfig.java
bemasc: OK, done.  I also added comments in the places where it would need to be initialized.

PR #260, opened 2020-04-17
Re-enable RemoteConfig fetches

bemasc:


Android/app/src/main/java/app/intra/sys/AutoStarter.java
```diff
@@ -47,9 +48,8 @@ public void onReceive(Context context, Intent intent) {
         context.startActivity(startIntent);
         return;
       }
-      controller.start(context);
-      // If RemoteConfig is being used, delay the above line until after a call to
-      // RemoteConfig.update() completes.
+      // Delay start until after the remote configuration has been updated, or failed to update.
+      RemoteConfig.update().addOnCompleteListener(success -> controller.start(context));
```
alalamav, line 52: Do we (want to) log the failure?

bemasc: Firebase logs it to the local logs, but we don't currently try to report it back to Firebase.  Doing that seems like something of a paradox: if we can't reach Firebase to fetch the remote configuration (which can certainly happen if the network isn't ready), we probably can't reach Firebase to report the failure either.

PR #264, opened 2020-05-01
Update IP script to Python3

bemasc:
Also includes updated IP databases for May 2020


scripts/dbip_shrink.py
```diff
@@ -11,18 +12,17 @@
...
+infile = gzip.open(sys.argv[1], mode='rt')
 out_prefix = sys.argv[2]
 
 v4file = open(out_prefix + '.v4', 'wb')
 v6file = open(out_prefix + '.v6', 'wb')
 
-r = csv.reader(infile)
-for start, end, country in r:
-  a = ipaddress.ip_address(unicode(start))
+for start, end, country in csv.reader(infile):
```
alalamav, line 21: Since you're decompressing as text, should we specify the file encoding?

bemasc: I'm not sure I understand.  Are you suggesting a comment, or a different variable name?

alalamav: I was wondering if you had to call `csv.reader(infile, encoding='utf-8')`. If it's working as is, it may not be necessary.

bemasc: From the [docs](https://docs.python.org/3/library/csv.html#csv.reader), it looks like that isn't an option.  We could specify a "dialect", but since it appears to be working I'm not sure what to specify.

cjhenck: I think you specify encoding when opening the file: https://docs.python.org/3/library/csv.html#examples

PR #266, opened 2020-05-07
Switch to ViewPager2 for the intro screens

bemasc:
ViewPager is deprecated, and ViewPager2 has
some improvements, like RTL support.

### DIFF
--- a/Android/app/src/main/java/app/intra/ui/IntroDialog.java
+++ b/Android/app/src/main/java/app/intra/ui/IntroDialog.java
@@ -26,12 +26,16 @@ import android.widget.Button;
 import android.widget.ImageView;
 import android.widget.TextView;
 import androidx.annotation.DrawableRes;
+import androidx.annotation.NonNull;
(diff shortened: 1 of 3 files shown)


Android/app/src/main/java/app/intra/ui/IntroDialog.java
```diff
@@ -174,32 +176,29 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
     }
 
     @Override
-    public Fragment getItem(int position) {
-      if (position >= 0 && position < pages.length) {
-        return pages[position];
-      }
-      return null;
+    public @NonNull Fragment createFragment(int position) {
+      return pages[position];
```
alalamav, line 194: Is `position` guaranteed to stay in bounds?

bemasc: Yes, I believe it is guaranteed to be `0 ... getItemCount()-1`.

PR #269, opened 2020-05-30
Integrate Choir metrics

bemasc:


Android/app/src/main/java/app/intra/sys/firebase/AnalyticsWrapper.java
```diff
@@ -20,6 +20,8 @@
 import android.net.NetworkInfo;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
+import app.intra.sys.CountryCode;
```
alalamav, line 24: Are you using the import on this file?

bemasc: Yes.  I moved the `CountryCode` class out of this package, so an import is now required here.

PR #277, opened 2020-08-05
Upgrade to Firebase Crashlytics

bemasc:
This requires increasing the minSdkVersion to 16, which unblocks a
number of other dependency upgrades and allows us to remove some version
checks.


bemasc:
There's a lot of noise in this due to dependency tree upgrades, but you can hide them from the review with a file-type filter: https://github.com/Jigsaw-Code/Intra/pull/277/files?file-filters%5B%5D=.gradle&file-filters%5B%5D=.java&file-filters%5B%5D=.xml

PR #284, opened 2020-09-11
Mark the VPN as unmetered

bemasc:
Starting in Android Q (10), VPNs are marked as "metered" by default,
preventing low-priority tasks like backups from running while the VPN is
on.  Intra is not metered, so we have to mark it as such explicitly.


alalamav (approved):

jemalhemrayew:
Kod barmy

PR #366, opened 2020-11-14
Revert "Mark the VPN as unmetered"

shatlyktma:
https://dns.dnshome.de/dns-query


shatlyktma:
> https://dns.dnshome.de/dns-query

PR #406, opened 2020-12-02
Stub out CertificateLoader support

SeanBurford:
Requires Android/tun2socks/tun2socks.aar be updated with https://github.com/Jigsaw-Code/outline-go-tun2socks/pull/68 to support the API changes.


bemasc:
I rolled these changes into #410, so we should be ready to add client cert support on the Java side.

PR #410, opened 2021-01-11
Incorporate updated outline-go-tun2socks

bemasc:
This slightly changes the names used in the Go-Java interface
and adds new arguments for certificate authentication (not yet used).

### DIFF
--- a/Android/app/src/main/java/app/intra/net/go/GoIntraListener.java
+++ b/Android/app/src/main/java/app/intra/net/go/GoIntraListener.java
@@ -37,7 +37,7 @@ import split.RetryStats;
  * when a socket has concluded, with performance metrics for that socket, and this class forwards
  * those metrics to Firebase.
  */
-public class GoIntraListener implements tunnel.IntraListener {
+public class GoIntraListener implements intra.Listener {
 
   // UDP is often used for one-off messages and pings.  The relative overhead of reporting metrics
   // on these short messages would be large, so we only report metrics on sockets that transfer at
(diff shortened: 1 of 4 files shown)


bemasc:
This essentially the same as #406, but I wanted to make sure we were including the latest build of outline-go-tun2socks.  I was also surprised by the large drop in binary size in #406, and am similarly surprised by the _different_ reduction in size observed here.

In a basic smoketest, the build appears to be working.

fortuna (approved):

ignoramous:
We're seeing ~10MB reduction in size. Not sure what triggered that... but we were being very _lax_ with conflicts resulting from go.mod / go.sum changes.

We have since released it to 1000 odd testers (Android 6+), it has been 2 weeks and no one has reported any issues thus far.

bemasc:
Thanks for the update @ignoramous.

SeanBurford:
On the size difference:

libgojni.so, which is built and included for several architectures, accounts for the size difference.

objdump -h libgojni.so (x86) shows that:
*  The dynamic relocation table .rel.dyn has shrunk 378k
*  The executable section .text has shrunk 1,599k
*  Static data section .rodata has shrunk 327k
*  Relocatable static data section .data.rel.ro has shrunk 1,872k

Then, comparing the symbols we see that a lot of support libs (`diff --side-by-side /tmp/pre /tmp/post` using `objdump  -t libgojni.so | awk '/.text/{print $6, $5}'| sort > /tmp/post`) have been refactored out:

*  encoding/json
*  expvar
*  fmt.(*ss).*
*  github.com/cespare/xxhash
*  github.com/golang/protobuf
*  github.com/op/go-logging
*  github.com/prometheus/client_golang
*  shadowsocks2
*  golang.org/x/cryptobyte
*  mime/multipart

PR #416, opened 2021-01-29
Fix x86-64 compatibility and reduce size

bemasc:
The previous version was built with the BoringCrypto branch of Go, which
produces binaries that are larger and are incompatible with x86-64
(https://github.com/golang/go/issues/35441).


alalamav (approved):

SaintPatricksCoin (approved):
@SaintPatricksCoin

Nexuscompute (commented):
**__**

PR #425, opened 2021-12-17
Switch to a modern dependency locking system

bemasc:
This change removes the previous supply-chain defense, requiring all
third-party dependencies to be bundled into the git repo.  Instead, the
build files enable Gradle's new-ish support for locking and hashing
dependencies, ensuring that the build is reproducible without requiring
all dependencies to be checked in.


cjhenck (approved):
LGTM

PR #428, opened 2022-02-13
Add AdGuard Public DNS

PeterDaveHello:
AdGuard is also a popular DNS. Would like to see more options on the list 🎉

Reference:

- https://adguard-dns.com/en/public-dns.html


PeterDaveHello:
@bemasc sorry that I didn't notice, not sure if there's any reason behind this decision that can be shared? Many thanks ;)

PR #429, opened 2022-02-25
Update 42l IP addresses

42l-neil-tmp:
Hi,

We migrated our servers, we have a new IPv4 address and we now support IPv6. Here’s a PR to reflect those changes in Intra.


42l-neil-tmp:
I accept the CLA, but signing it requires me to connect to a Google account. I don’t have any and I don’t plan to create one.

42l-neil-tmp:
Hi,

This is getting urgent. On May 1st, our old server will stop working and this IP address won’t answer to your queries anymore. **It may impact your users** − you’ve been warned.

The only thing you folks need to do is to update our server’s IP address by merging this PR.

PR #436, opened 2022-04-28
Update tun2socks.aar

bemasc:
This change also adjusts the persistent notification flags as
required by Android S.


fortuna (approved):
Great! I guess we need to do something similar for Outline.

PR #485, opened 2023-09-01
feat: add Backend that uses outline-sdk to replace tun2socks

jyyi1:
**Actions**

- [x] Implementing Intra StreamDialer and PacketProxy
- [x] Use Gradle to invoke gomobile
- [x] Apply changes to go-tun2socks first: https://github.com/Jigsaw-Code/outline-go-tun2socks/pull/125
- [x] Move go-tun2socks code to this repository (#487)
- [ ] Refactor folder structure


go.mod
```diff
@@ -0,0 +1,25 @@
+module github.com/Jigsaw-Code/Intra
+
+go 1.20
+
+require (
+	github.com/Jigsaw-Code/choir v1.0.1
+	github.com/Jigsaw-Code/outline-go-tun2socks v0.0.0-20230815223204-dada2652ae2c
```
fortuna, line 7: Why do we need this?

Android/backend/intra/internal/doh/packet_proxy.go
```diff
@@ -0,0 +1,196 @@
...
+import (
+	"errors"
+	"fmt"
+	"log"
+	"net"
+	"net/netip"
+	"sync/atomic"
+	"time"
+
+	intraLegacy "github.com/Jigsaw-Code/outline-go-tun2socks/intra"
```
fortuna, line 26: Let's move the code we need to this repo instead of depending on the old code.

Android/backend/tools.go
```diff
@@ -0,0 +1,25 @@
...
+//go:build tools
+// +build tools
+
+// See https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+// and https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md
+
+package tools
+
+import (
+	_ "golang.org/x/mobile/bind"
```
fortuna, line 24: This is technique is for binaries, not libraries.

You need the gomobile and gobind binaries. See my outline-sdk pr

Android/backend/intra/internal/doh/packet_proxy.go
fortuna, line 1: Let's put IP-level stuff in its own package. Perhaps name it `vpn`?

Android/backend/intra/tun_unix.go
fortuna, line 1: Move to the vpn library

Android/backend/intra/device.go
```diff
@@ -0,0 +1,104 @@
...
+
+	log.Println("[info] Intra device initialized")
+	return
+}
+
+func (d *IntraDevice) Close() error                { return d.t2s.Close() }
+func (d *IntraDevice) Read(p []byte) (int, error)  { return d.t2s.Read(p) }
+func (d *IntraDevice) Write(p []byte) (int, error) { return d.t2s.Write(p) }
+
+func (d *IntraDevice) UpdateDoHServer(serverURL, fallbackAddrs string) error {
```
fortuna, line 87: What does this do? When is this needed?

Android/backend/intra/internal/doh/config.go
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package doh
+
+import "strings"
+
+func parseFallbackAddrs(ipList string) []string {
```
fortuna, line 19: Perhaps move this to transport.go, since it's only used there?

fortuna (changes requested):
It's hard to review this code without the corresponding Android change, and the build code.
It's not clear what is actually consumed by Android.

I need to clearly see what the interface between Android and Go is. It seems like:
- Tun2socks.connectIntraTunnel
- Tun2socks.newDoHTransport

What else? Can you put the entire interface used somewhere we can discuss?

Let's move the backend code out of Android, so it's clear it could be used by iOS if we add support.  But also to make the boundary clear.
We should rename it though. "Backend" is ambiguous.

Perhaps we put the Go code in some `/go/internal` root dir. The Go backend would be a specific package (or packages) in that directory. Perhaps we call it `/go/internal/intrago` or `/go/internal/intraapi` or `/go/internal/gobackend`. Let's discuss.

The doh code shouldn't be nested in the backend, since that's code that perhaps could be migrated to the SDK one day.

fortuna (changes requested):
Perhaps to minimize change, we just move everything we need from outline-go-tun2socks into /go/internal/..., then make the minimal changes needed to use the SDK network APIs.

We should have no dependencies on outline-go-tun2socks.

I suggest PR #1 is just to move the code here, and figure out the build.

Then PR #2 can actually make use the SDK network apis.

Android/backend/tools.go
jyyi1: `gomobile` still needs this package. Otherwise it will report missing package error.

jyyi1:
@fortuna No I removed all old interfaces, and only expose types in `backend/intra` package. Just pushed the Java change to this repo. I didn't figure out a good way to call `gomobile` using gradle (there was an official plugin, but it was deprecated long time ago), so you still need to call `gomobile` in terminal (as mentioned in `backend/README.md`), and then build the Android project using gradle.

Android/backend/README.md
```diff
@@ -0,0 +1,10 @@
+# Intra Backend
+
+Intra backend is a Go project using Outline SDK.
+
+## Build
+
+```sh
+gomobile clean
```
fortuna, line 8: It's not clear how to set up go mobile.
We need to specify it in tools.go to make sure the version is fixed.
You shouldn't use the system binary.

Android/backend/tools.go
fortuna: The best approach I've found is to build both binaries locally to avoid that issue.
See https://github.com/Jigsaw-Code/outline-sdk/tree/fortuna-appproxy/x/examples/mobileproxy
Notice the tools.go file too.

Unfortunately gomobile calls gobind as a binary, instead of a library, so it must be on the PATH to be found. That's why the PATH trick there.

fortuna (changes requested):

Android/backend/README.md
```diff
@@ -0,0 +1,10 @@
...
+gomobile bind -target=android -o ./Android/backend/backend.aar ./Android/backend/intra
```
fortuna, line 9: I see the api is all in one package. Nice.

Can you move it so that there aren't things nested in it?
I was imagining something like

- go/
  - backend/ (or some name that is clearer)
  - internal/
    - doh/
    - sni/
    - vpn/  (we should move the vpn stuff out of doh)

Android/backend/intra/tun_unix.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// MakeTunFile returns an os.File object from a TUN file descriptor `fd`.
+// The returned os.File holds a separate reference to the underlying file,
+// so the file will not be closed until both `fd` and the os.File are
+// separately closed.  (UNIX only.)
+func NewTunFromFile(fd int) (IPDevice, error) {
```
fortuna, line 28: `NewDeviceFromFileDescriptor` or `NewDeviceFromFD`?

Android/backend/intra/device.go
fortuna: Found it.

Android/backend/intra/device.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/Jigsaw-Code/outline-sdk/network/lwip2transport"
+)
+
+// SocketProtector is a way to make certain sockets or DNS lookups bypassing the VPN connection. This is only needed
+// for devices running Android versions older than Lollipop (21). Once a socket is protected, data sent through it will
+// go directly to the internet, bypassing the VPN. The Android VpnService implements the protect() method.
+type SocketProtector doh.Protector
+
+type IntraDevice struct {
+	t2s       network.IPDevice
```
fortuna, line 36: Since you are just extending the IPDevice, you can do this, so you don't need to specify all method wrappers:
```suggestion
type IntraDevice struct {
       network.IPDevice
```

Android/backend/intra/internal/doh/stream_dialer.go
```diff
@@ -0,0 +1,134 @@
...
+	dohServer        atomic.Pointer[DoHTransport]
+	dialer           *net.Dialer
+	alwaysSplitHTTPS atomic.Bool
+	listener         intraLegacy.TCPListener
+	sniReporter      sni.TCPSNIReporter
+}
+
+var _ DoHStreamDialer = (*dohSplitStreamDialer)(nil)
+
+func MakeDoHStreamDialer(fakeDNS netip.AddrPort, dohServer DoHTransport, protector Protector, listener intraLegacy.TCPListener, sniReporter sni.TCPSNIReporter) (DoHStreamDialer, error) {
```
fortuna, line 54: This isn't a DoH StreamDialer. It's an `IntraStreamDialer`. It does a lot more than DoH. We need to rename it.

Android/backend/intra/internal/doh/transport.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package doh
+
+import (
+	dohLegacy "github.com/Jigsaw-Code/outline-go-tun2socks/intra/doh"
+	"github.com/Jigsaw-Code/outline-go-tun2socks/intra/protect"
+)
+
+type DoHTransport = dohLegacy.Transport
```
fortuna, line 22: These are really hurting readability. We should move all the code here.

fortuna (changes requested):

fortuna (changes requested):
There are two main issues with this PR:
- Duplication of definitions
- Not clear what's original, and what's changing

It makes it really hard to review because you are moving some code, but not all, and changing it at the same time.

To make this more understandable, can you split the PR, so the move and the change are separate and clear? You can move then change, or change then move.

The move shouldn't change code. Moving to Intra first would give us the benefit of being able to build everything in one repo, though it seems we would still depend on some dependencies shared with Outline in outline-go-tun2socks, which is bad.

Perhaps it's best to change in outline-go-tun2socks first, removing all dependencies on legacy code, then move to Intra. The code in Intra should no longer depend on outline-go-tun2socks, and there shouldn't be duplication of definitions in the new code.

Android/app/src/main/java/app/intra/net/go/GoProber.java
```diff
@@ -38,22 +37,13 @@ public GoProber(Context context) {
   @Override
   public void probe(String url, Callback callback) {
     new Thread(() -> {
-      String dohIPs = GoVpnAdapter.getIpString(context, url);
+      String dohIPs = GoVpnAdapter.getDoHServerFallbackIPsString(context, url);
       try {
         // Protection isn't needed for Lollipop+, or if the VPN is not active.
-        Protector protector = VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP ? null :
+        SocketProtector protector = VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP ? null :
```
fortuna, line 43: Let's get rid of the protector, it complicates things and is no longer needed.

I took a look in the Play Store, and only 0.2% of the installed audience is below Lollipop (Android 5): https://docs.google.com/spreadsheets/d/15zsfgMhJ7TsLumE6S5ZykETwMBvrBptFllX0Z02gFPw/edit?usp=sharing&resourcekey=0-inA8jaG0XS58zXNXU9YOzQ

The global stats say 0.66%: https://apilevels.com/

It's not worth the trouble.

Let's make that change separately though. Change, then move.

fortuna (changes requested):

Android/app/src/main/java/app/intra/net/go/GoProber.java
fortuna: As we remove the protector, we should also update the minSdkVersion to 21 (Lollipop), to make sure older phones don't get the new version that will break.

intra882:
Дайте пожалуйста код 
--
Отправлено из Mail.ru для Android среда, 06 сентября 2023г., 12:51 +05:00 от Vinicius Fortuna  ***@***.*** :

Android/backend/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+
+task ensureBuildDir() {
+    doLast {
+        goBuildDir.mkdirs()
+    }
+}
+
+task ensureGoMobile(type: Exec, dependsOn: ensureBuildDir) {
+    environment 'GOBIN', goBuildDir.getPath()
+    commandLine 'go', 'install', 'golang.org/x/mobile/cmd/gomobile@latest'
```
fortuna, line 29:
```suggestion
    commandLine 'go', 'install', 'golang.org/x/mobile/cmd/gomobile@latest', 'golang.org/x/mobile/cmd/gobind@latest'
```

Android/backend/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+} doLast {
+    exec {
+        environment 'GOBIN', goBuildDir.getPath()
+        commandLine "${goBuildDir}/gomobile", 'init'
+    }
+}
```
fortuna, line 35:
```suggestion
}
```

Android/backend/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+
+task gobind(type: Exec, dependsOn: ensureGoMobile) {
+    workingDir goBuildDir
+    environment 'ANDROID_HOME', android.sdkDirectory
+    environment 'PATH', goBuildDir.getPath() + System.getProperty("path.separator") + System.getenv("PATH")
+    commandLine "${goBuildDir}/gomobile", 'bind', '-target=android', '-o', outputAAR, "${rootDir}/backend/intra"
```
fortuna, line 41: It would be nice to specify the `-androidapi` here.

jyyi1:
This has been partially implemented in a bunch of other PRs, such as #496, #501 . Since there has been already so many conflicts , it would be better to just create new PRs continuing the work instead of modifying this one.

PR #498, opened 2024-01-31
fix: crash in work profile due to exceptions thrown by TileService

jyyi1:
I fixed a crash issue caused by: [`TileService.requestListeningState`](https://developer.android.com/reference/android/service/quicksettings/TileService#requestListeningState(android.content.Context,%20android.content.ComponentName)) throwing exceptions if the user of the context is not the current user (might be a work profile user).

This is the new behavior since API level 33, so previously this issue is not found.


fortuna (approved):
Thanks for the fix.

PR #506, opened 2024-03-13
refactor: remove unused `alwaysSplitHTTPS` flag

jyyi1:
The code that modifies the `alwaysSplitHTTPS` flag has been removed in #241 . So now it will always be `false`, and there is no need to keep that branch any more.


fortuna (approved):
Thanks for the cleanup.

PR #509, opened 2024-04-04
Create ant.yml

Laiiev:


.github/workflows/ant.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up JDK 11
+      uses: actions/setup-java@v3
+      with:
+        java-version: '11'
+        distribution: 'temurin'
+    - name: Build with Ant
+      run: ant -noinput -buildfile build.xml
```
jyyi1, line 25: What is `build.xml` here? I don't think we have that file.

jyyi1 (changes requested):
Thanks. But it would be better to add a gradle CI job here, which should cover the Java part.

PR #516, opened 2024-06-25
split: retrier.conn must be guarded by mutex

ignoramous:


ignoramous:
cc: @bemasc / @jyyi1

jyyi1 (approved):
Thanks for the contribution!

PR #517, opened 2024-06-25
split: copyOnce may be called more than once

ignoramous:
Called twice when `retrier.Write` succeeds and returns immediately (L319) without waiting on `<-retryCloseFlag` (L324).

https://github.com/Jigsaw-Code/Intra/blob/5c6d68dec9782c89a53fb3f37366299c0192d240/Android/app/src/go/intra/split/retrier.go#L317-L324

### DIFF
--- a/Android/app/src/go/intra/split/retrier.go
+++ b/Android/app/src/go/intra/split/retrier.go
@@ -351,11 +351,11 @@ func copyOnce(dst io.Writer, src io.Reader) (int64, error) {
 
 func (r *retrier) ReadFrom(reader io.Reader) (bytes int64, err error) {
 	for !r.retryCompleted() {
-		var b int64
-		if b, err = copyOnce(r, reader); err != nil {
+		n, e := copyOnce(r, reader)
+		bytes += n
+		if err = e; err != nil {
 			return
 		}
-		bytes += b
 	}
 
 	var b int64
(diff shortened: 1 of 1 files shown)


Android/app/src/go/intra/split/retrier.go
```diff
@@ -351,9 +351,11 @@ func copyOnce(dst io.Writer, src io.Reader) (int64, error) {
 
 func (r *retrier) ReadFrom(reader io.Reader) (bytes int64, err error) {
 	for !r.retryCompleted() {
-		if bytes, err = copyOnce(r, reader); err != nil {
+		var b int64
+		if b, err = copyOnce(r, reader); err != nil {
 			return
 		}
+		bytes += b
```
jyyi1, line 358: `b` might be `> 0` even when `err != nil`, so we should always add it to `bytes` regardless of the `err`.

```suggestion
		n, e := copyOnce(r, reader)
		bytes += n
		if err = e; err != nil {
			return
		}
```

jyyi1 (changes requested):

Android/app/src/go/intra/split/retrier.go
ignoramous: Done. Thanks.

jyyi1 (approved):
Thanks for the contribution!

PR #518, opened 2024-06-25
retrier: use atomics as flags

ignoramous:
toctou race in checking closed(ch) to actually closing it, which may result in a panic if chan ch is closed twice.

### DIFF
--- a/Android/app/src/go/intra/split/retrier.go
+++ b/Android/app/src/go/intra/split/retrier.go
@@ -62,8 +62,8 @@ type retrier struct {
 	// Flag indicating when retry is finished or unnecessary.
 	retryCompleteFlag chan struct{}
 	// Flags indicating whether the caller has called CloseRead and CloseWrite.
-	readCloseFlag  chan struct{}
-	writeCloseFlag chan struct{}
+	readCloseFlag  atomic.Bool
+	writeCloseFlag atomic.Bool
 	stats          *RetryStats
 }
 
@@ -87,11 +87,11 @@ func closed(c chan struct{}) bool {
 }
 
 func (r *retrier) readClosed() bool {
-	return closed(r.readCloseFlag)
+	return r.readCloseFlag.Load()
 }
 
 func (r *retrier) writeClosed() bool {
-	return closed(r.writeCloseFlag)
+	return r.writeCloseFlag.Load()
 }
 
 func (r *retrier) retryCompleted() bool {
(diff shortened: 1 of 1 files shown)


ignoramous:
- https://github.com/celzero/firestack/issues/64

Android/app/src/go/intra/split/retrier.go
```diff
@@ -211,12 +209,12 @@ func (r *retrier) retry(buf []byte) (n int, err error) {
 }
 
 func (r *retrier) CloseRead() error {
-	if !r.readClosed() {
-		close(r.readCloseFlag)
+	if r.readCloseFlag.CompareAndSwap(false, true) {
+		r.mutex.Lock()
+		defer r.mutex.Unlock()
+		return r.conn.CloseRead()
```
jyyi1, line 215: Please preserve the current behavior of forwarding multiple `CloseRead` calls to `r.conn`. This is useful when the first `r.conn.CloseRead()` returns an `error`, and the caller tries to call `CloseRead()` again later, we should forward the second call to `r.conn`. So we may simply need a `r.readCloseFlag.Store(true)` here.

jyyi1 (commented):
Thanks for the contribution! The code LGTM, with some minor comments.

Android/app/src/go/intra/split/retrier.go
ignoramous: thanks! done.

jyyi1 (approved):

Android/app/src/go/intra/split/retrier.go
```diff
@@ -62,8 +62,8 @@ type retrier struct {
 	// Flag indicating when retry is finished or unnecessary.
 	retryCompleteFlag chan struct{}
 	// Flags indicating whether the caller has called CloseRead and CloseWrite.
-	readCloseFlag  chan struct{}
-	writeCloseFlag chan struct{}
+	readCloseFlag  atomic.Bool
```
jyyi1, line 65: Please `import "atomic"` package. (see the build failure below)

jyyi1 (changes requested):

Android/app/src/go/intra/split/retrier.go
ignoramous: yikes!

jyyi1 (approved):

PR #519, opened 2024-06-26
split: skip zero len writes

ignoramous:
I'm not sure why (as zero len writes are apparently no-op in Golang) but the Quad9 resolver (DNSCrypt over TCP) doesn't work without this check.

### DIFF
--- a/Android/app/src/go/intra/split/retrier.go
+++ b/Android/app/src/go/intra/split/retrier.go
@@ -328,11 +328,12 @@ func (r *retrier) Write(b []byte) (int, error) {
 			r.mutex.Lock()
 			r.mutex.Unlock()
 	
-			if len(leftover) > 0 {
-				m, err := r.conn.Write(leftover)
-				return n + m, err
+			if buf := b[n:]; len(buf) > 0 {
+				m, e := r.conn.Write(buf)
+				n += m
+				err = e
 			}
-			return n, nil
+			return n, err
 		}
 	}
 
--- a/Android/app/src/go/intra/split/retrier.go
+++ b/Android/app/src/go/intra/split/retrier.go
@@ -324,12 +324,13 @@ func (r *retrier) Write(b []byte) (int, error) {
 			<-r.retryCompleteFlag
 	
 			r.mutex.Lock()
+			c := r.conn
 			r.mutex.Unlock()
 
 			// zero len writes are no-ops, but Quad9 servers
 			// are observed to respond better when these are skipped.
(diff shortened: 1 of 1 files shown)


Android/app/src/go/intra/split/retrier.go
```diff
@@ -318,14 +318,21 @@ func (r *retrier) Write(b []byte) (int, error) {
...
 			r.mutex.Lock()
 			r.mutex.Unlock()
-			m, err := r.conn.Write(b[n:])
-			return n + m, err
+	
+			if len(leftover) > 0 {
+				m, err := r.conn.Write(leftover)
+				return n + m, err
+			}
+			return n, nil
```
jyyi1, line 335:
```suggestion
			if buf := b[n:]; len(buf) > 0 {
				m, e := r.conn.Write(buf)
				n += m
				err = e
			}
			return n, err
```

Android/app/src/go/intra/split/retrier.go
```diff
@@ -318,14 +318,21 @@ func (r *retrier) Write(b []byte) (int, error) {
 			if err == nil {
 				return n, nil
 			}
+
+			leftover := b[n:]
```
jyyi1, line 322:
```suggestion
```

Android/app/src/go/intra/split/retrier.go
```diff
@@ -318,14 +318,21 @@ func (r *retrier) Write(b []byte) (int, error) {
...
 			// A write error occurred on the provisional socket.  This should be handled
 			// by the retry procedure.  Block until we have a final socket (which will
 			// already have replayed b[:n]), and retry.
 			<-r.retryCompleteFlag
+	
 			r.mutex.Lock()
 			r.mutex.Unlock()
-			m, err := r.conn.Write(b[n:])
-			return n + m, err
+
```
jyyi1, line 330: Please also add a comment here to explain the reason we are skipping 0-len buffers.

jyyi1 (changes requested):

Android/app/src/go/intra/split/retrier.go
ignoramous: Nice. Thanks.

jyyi1 (approved):

Kjs7945:
> I'm not sure why (as zero len writes are apparently no-op in Golang) but the Quad9 resolver (DNSCrypt over TCP) doesn't work without this check.

dekwan44 (approved):

PR #520, opened 2024-08-16
retrier: confirm splitHello to io.Writer.Write

ignoramous:
The way our Intra fork is written, writing to `retrier` (`DuplexConn`) via `io.Copy` results in panic when total written bytes from TLS fragmentation exceeds the length of the input bytes; expected since this behaviour violates the [`io.Writer.Write`](https://pkg.go.dev/io#Writer) contract:

> It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. Write must return a non-nil error if it returns n < len(p).

https://github.com/celzero/firestack/commit/ba0c57c3851ce8c08c2f392b57a1890a936c803c

### DIFF
--- a/Android/app/src/go/intra/split/retrier.go
+++ b/Android/app/src/go/intra/split/retrier.go
@@ -283,9 +283,11 @@ func splitHello(hello []byte, w io.Writer) (n int, splitLen int, err error) {
 	pkt = hello[splitLen-5:]
 	copy(pkt, hello[:5])
  	binary.BigEndian.PutUint16(pkt[3:5], recordLen-uint16(recordSplitLen))
-	var m int
-	m, err = w.Write(pkt)
-	n += m
+	_, err = w.Write(pkt)
+
+	// total written bytes after fragmentation will always be greater than the length
+	// of the input buffer, which may break clients expecting to not exceed it.
+	n = len(hello)
 	return
 }
 


jyyi1 (changes requested):
Returning `len(hello)` seems simple, but calculating actual written bytes is trickier, especially if a `Write` fails. We may or may not subtract the header length from `n`, depending on the bytes that has been written.

Since this is a private function, and we are not using the returned bytes, I recommend leaving it unchanged.

Please do not use this function elsewhere. For reusable TLS fragmentation, use the [Outline SDK](https://pkg.go.dev/github.com/Jigsaw-Code/outline-sdk@v0.0.16/transport/tlsfrag).

ignoramous:
> Since this is a private function, and we are not using the returned bytes

`splitter` (which impls `io.Writer` via `DuplexConn`) does use `n` returned from `splitHello`.

https://github.com/Jigsaw-Code/Intra/blob/27637e0ed497dbd1e862750b14f9a5c0be368f15/Android/app/src/go/intra/split/direct_split.go#L57

> Please do not use this function elsewhere. For reusable TLS fragmentation, use the Outline SDK. (We plan to refactor our code using the Outline SDK in the future as well)

Ah, thanks.

> We may or may not subtract the header length from n, depending on the bytes that has been written ... I recommend leaving it unchanged

I didn't quite catch this. 

You reckon we change `splitter.Write` to confirm to `io.Writer` or close this PR?

jyyi1:
Thanks @ignoramous for your response. I suggest closing this PR if its purpose is solely for third-party use (Outline SDK is a more suitable API with proper handling of written bytes), since we'll be refactoring this logic anyway.

However, if you're aware of any Intra use cases that might be affected, we may need to fix it.

ignoramous:
> we'll be refactoring this logic anyway.

Gotcha.

> However, if you're aware of any Intra use cases that might be affected, we may need to fix it.

Well, only `splitter` (aka `direct_split`) but it isn't used by Intra anymore, even though (as it was brought to my attention, `direct_split` works better in face of ["residual censorship"](https://datatracker.ietf.org/doc/html/rfc9505#name-residual-censorship)).

(thanks for your input. closing...)

PR #524, opened 2024-10-31
Upgrade to AGP 8.5.2

mzennis:
### Changes
- [x] upgrade to AGP 8.5.2 (compatible with Android Studio Koala | 2024.1.1 Patch 2)
- [x] upgrade Gradle wrapper to 8.7 (comply with https://developer.android.com/build/releases/gradle-plugin)
- [x] upgrade Firebase to latest version
- [x] set minimum sdk to 21
- [x] disable android.nonTransitiveRClass
- [x] disable android.nonFinalResIds


Uuptupyy:
(
 (http.host eq "api.example.com" and http.request.uri.path eq "/api/v2/auth") or
 (http.host matches "^(www|store|blog)\.example\.com" and http.request.uri.path contains "wp-login.php") or
 ip.geoip.country in {"CN" "TH" "US" "ID" "KR" "MY" "IT" "SG" "GB"} or ip.geoip.asnum in {12345 54321 11111}
) and not ip.src in {11.22.33.0/24}

Android/build.gradle
```diff
@@ -1,7 +1,7 @@
-dependencyLocking {
-    lockAllConfigurations()
```
jyyi1, line 2: We still need to lock the dependencies, this might lead to a large amount of changes.

Amirjobi (approved):

Android/build.gradle
mzennis: hi @jyyi1 please help to re-review my PR, I was adding two more commits:
- https://github.com/Jigsaw-Code/Intra/pull/524/commits/a35b414de5d68d2aa1069dff828186000bff6dc7 
- https://github.com/Jigsaw-Code/Intra/pull/524/commits/cfe5ca137f86907ab6d2290d30d14dcfff7c6cc6

let me know if you have any feedback

Android/app/build.gradle
```diff
@@ -125,11 +129,12 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:2.1.2'
     implementation "androidx.viewpager2:viewpager2:1.0.0"
     // For Firebase Analytics, etc.
-    implementation 'com.google.firebase:firebase-analytics:19.0.2'  // Last version for API <19
-    implementation 'com.google.firebase:firebase-perf:20.0.4'
-    implementation 'com.google.firebase:firebase-crashlytics:18.2.6'
-    implementation 'com.google.firebase:firebase-crashlytics-ndk:18.2.6'
-    implementation 'com.google.firebase:firebase-config:21.0.1'
+    implementation 'com.google.firebase:firebase-analytics:22.1.2'  // Last version for API <19
```
jyyi1, line 132: I think this comment is not accurate any more?

```suggestion
    implementation 'com.google.firebase:firebase-analytics:22.1.2'
```

Android/app/build.gradle
```diff
@@ -94,6 +95,9 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    buildFeatures {
+        buildConfig = true
+    }
```
jyyi1, line 100: Are we using it anywhere in the code? If not, I think we can remove this.

Android/gradle.properties
```diff
@@ -25,5 +25,8 @@ android.useAndroidX=true
 ANDROID_COMPILE_SDK_VERSION=33
 ANDROID_BUILD_TOOLS_VERSION=33.0.0
 
-ANDROID_MIN_SDK_VERSION=16
+ANDROID_MIN_SDK_VERSION=21
 ANDROID_TARGET_SDK_VERSION=33
+
+android.nonTransitiveRClass=false
+android.nonFinalResIds=false
```
jyyi1, line 32: Are these two settings necessary?

jyyi1 (commented):
Thanks for the update. I have left some comments.

Fakrul877 (commented):
Bagus dan terima kasih untuk amati segalanya mungkin tak bisa saya dalami dan saya juga pengguna baru  tolong lah beri tunjuk ajar

Android/app/build.gradle
mzennis: yes, we are actually using it (code: https://github.com/Jigsaw-Code/Intra/blob/5a48272f0ba74e12b3f27fd3c5d82e0a0b240457/Android/app/src/main/java/app/intra/sys/firebase/LogWrapper.java#L19)

but, we could change the implementation, instead of using `BuildConfig`. like, differentiate `LogWrapper` for release and debug (`src/debug/`) mode. wdyt?

Android/gradle.properties
mzennis: well, the `android.nonTransitiveRClass=false` is because, we are using: 
Screenshot 2025-01-24 at 22 00 19

Also, I don't think we need to change how it's done, since I'm worried it might mess up other stuff.

mzennis: my suggestion is, if we're going to change these, we should do it in a separate PR. wdyt?

jyyi1:
FYI, I have created #544 to only upgrade AGP to the minimum viable version, this should unblock most of the work flows.

mzennis:
> FYI, I have created #544 to only upgrade AGP to the minimum viable version (without too much dependency update), this should unblock most of the work flows.

thank you! I will close the PR then 👍🏻

PR #544, opened 2025-04-01
build: upgrade AGP to 8.1.4

jyyi1:
- **AGP Upgrade (7.0.4 -> 8.1.4):** Fixes an R8 bug in the older AGP that caused build failures with recent `gomobile` binaries. Version 8.1.4 includes the fix and supports target SDK 33.
- **Dependency Updates:** Updated `google-services` and Firebase dependencies to be compatible with AGP 8.1.4.
- **Plugin Fix:** Fixed the `google-services` plugin in `build.gradle` (moved to end, removed duplicate).


fortuna (approved):
Thanks!

miladsobhanii (approved):

PR #554, opened 2025-05-27
Remove some non-determinism

licaon-kter:
Your app is not built reproducible in F-Droid, but we continuously test older versions on https://verification.f-droid.org/ and would like more and more apps to become repro

Looking at the latest app report: https://verification.f-droid.org/packages/app.intra/

We can fix the GO BUILDID and the different /tmp paths.

### DIFF
--- a/Android/app/build.gradle
+++ b/Android/app/build.gradle
@@ -157,8 +157,9 @@ tasks.register('compileGoBackend', Exec) {
                         System.getenv('PATH')
 
     commandLine("${goBuildDir}/gomobile", 'bind',
-                '-ldflags=-s -w',
+                '-ldflags=-s -w -buildid=',
                 '-target=android',
+                '-trimpath',
                 "-androidapi=${android.defaultConfig.minSdk}",
                 '-o', goBackendAAR,
                 *goSourcePackages)


licaon-kter:
Haha CLA, no, but you can take this as you please :wink:

jyyi1:
Thanks for the contribution. But unfortunately this PR won't make the app reproducible. Because to ensure a reproducible build, we need to use a fixed version of `gomobile`, but they only officially provide the latest [`0.0.0` version](https://pkg.go.dev/golang.org/x/mobile/cmd/gomobile?tab=versions). This means that even if we fix the `BUILDID`, the produced binary isn't guaranteed to be reproducible.

licaon-kter:
True, but I've tested this, and as the title says it's about "some" of it. :)

Could this be made perfect? I hope so, in the mean time, this PR improves things.

jyyi1 (approved):
Thanks. That sounds reasonable to me. Please [accept the CLA](https://github.com/Jigsaw-Code/Intra/pull/554/checks?check_run_id=42976900928) so we can merge the PR.

licaon-kter:
As said, I won't submit to any CLA. :)

ignoramous:
> ... to ensure a reproducible build, we need to use a fixed version of gomobile, but they only officially provide ... 0.0.0 ...

Also, `gomobile` driven builds don't yet respect `-trimpath`, at least not from what we've observed (from build of our Intra fork).

- https://github.com/golang/go/issues/40254

miladsobhanii (approved):

Muhh96 (approved):

Android/app/build.gradle
```diff
@@ -157,8 +157,9 @@ tasks.register('compileGoBackend', Exec) {
                         System.getenv('PATH')
 
     commandLine("${goBuildDir}/gomobile", 'bind',
-                '-ldflags=-s -w',
+                '-ldflags=-s -w -buildid=',
```
fortuna: I believe the buildid may be used for caching and build tools, so there's some downside in removing it.

For a deterministic build, it seems we can simply overwrite it with https://pkg.go.dev/cmd/buildid

PR #607, opened 2026-06-22
release: 🎉 Intra v1.5.0

jyyi1:
This PR releases **Intra v1.5.0**.

- Bumped `versionName` from `1.4.1` to `1.5.0`.
- Trusted `*-sources.jar` and `*-javadoc.jar` in `verification-metadata.xml` to prevent Android Studio sync failures caused by the IDE downloading library sources as [suggested here](https://developer.android.com/build/dependency-verification#android_studio_synchronization).


Android/gradle/verification-metadata.xml
```diff
@@ -3,6 +3,11 @@
    <configuration>
       <verify-metadata>true</verify-metadata>
       <verify-signatures>false</verify-signatures>
+      <trusted-artifacts>
```
fortuna, line 6: Please add a link to the source of this hack.

fortuna (approved):

Android/gradle/verification-metadata.xml
eusgoogle: NAD. GŁOWY REBI REBEKA EXYN 00

PR #604, opened 2026-06-16
build: support 16KB page size and bump minSdkVersion to 21

jyyi1:
Aligns binary to 16 KB page-size for Android 15+ devices as [required by Google Play](https://developer.android.com/guide/practices/page-sizes#play-console).

#### Changes

- **Go Backend**: Added [C linker flags (`-extldflags=-Wl,-z,max-page-size=16384`)](https://developer.android.com/guide/practices/page-sizes#compile-r27-lower) to align `libgojni.so` to 16 KB.
- **Crashlytics**: Upgraded `firebase-crashlytics-ndk` to `19.0.2` to import 16 KB aligned native binaries (see [Crashlytics NDK v19.0.2 Release Notes](https://firebase.google.com/support/release-notes/android#crashlytics-ndk_v19-0-2)).
- **minSdkVersion**: Raised to 21. This bump is required by the upgraded Crashlytics SDK (see [Crashlytics v19.0.0 Release Notes](https://firebase.google.com/support/release-notes/android#crashlytics_v19-0-0)).
  - *Impact Metrics (as of June 16, 2026, via Google Play Store)*: Affects <0.013% (162 / 1,255,702) of legacy Android 4.1-4.4 users.
- **Metadata**: Updated `verification-metadata.xml` checksums.

PR #623, opened 2026-09-15
build(android): bump compile and target SDK to version 36

fortuna:
## What

Bumps the Android build to API level 36 (Android 16):

```diff
-ANDROID_COMPILE_SDK_VERSION=35
-ANDROID_BUILD_TOOLS_VERSION=35.0.0
+ANDROID_COMPILE_SDK_VERSION=36
+ANDROID_BUILD_TOOLS_VERSION=36.0.0

 ANDROID_MIN_SDK_VERSION=21
-ANDROID_TARGET_SDK_VERSION=35
+ANDROID_TARGET_SDK_VERSION=36
```

This mirrors #603 (the 34 -> 35 bump), which was also a `gradle.properties`-only change.

## Why no code changes are needed

I went through the Android 16 behavior changes that apply to apps *targeting* SDK 36
and checked each against this app:

| Behavior change | Applies to Intra? |
|---|---|
| Edge-to-edge opt-out (`windowOptOutEdgeToEdgeEnforcement`) removed | No — the app never used the opt-out, so it has been edge-to-edge since targetSdk 35. The root layouts already use `android:fitsSystemWindows="true"`. |
| Orientation / resizability restrictions ignored on large screens | No — the manifest declares no `screenOrientation`, `resizeableActivity` or `maxAspectRatio`. |
| Stricter `JobScheduler` quotas | No — the app uses a foreground service, not `JobScheduler`. |
| Predictive back defaults | No change — already the default at targetSdk 35. |
| 16 KB page alignment | Already handled in #604 / #609. |

## Verification

Run locally with JDK 17 and SDK platform 36 / build-tools 36.0.0:

- `./gradlew assembleDebug` — **BUILD SUCCESSFUL** (Go backend AAR rebuilt via `gomobile bind`)
- `./gradlew test` — **BUILD SUCCESSFUL**

## Follow-ups, now handled elsewhere in the stack

Bumping compileSdk to 36 surfaced one new lint error, `MissingPermission` for
`USE_FULL_SCREEN_INTENT` on the VPN-revoked notification. That turned out to be a real
bug rather than just a lint complaint, and is fixed in the second commit here.

AGP 8.5.1 also warned that it was only tested up to compileSdk 34. That is fixed by
#626, which upgrades AGP to 9.4.0.

Note that `./gradlew lintDebug` already fails on `master` with three pre-existing
errors (`StartActivityAndCollapseDeprecated`, `RestrictedApi`,
`QueryAllPackagesPermission`). This PR does not add to them. CI only runs `go test`.


peterwgnd (approved):
lgtm

PR #625, opened 2026-09-15
build: fix dependency verification on macOS and Windows

fortuna:
## What

Makes `./gradlew` work on macOS and Windows. Right now a clean checkout fails during
configuration, before any task runs:

```
> Dependency verification failed for configuration ':classpath'
  2 artifacts failed verification:
    - junit-bom-5.9.2.module (org.junit:junit-bom:5.9.2) from repository MavenRepo
    - junit-bom-5.9.3.module (org.junit:junit-bom:5.9.3) from repository MavenRepo
```

and then, a bit further in:

```
> Dependency verification failed for configuration ':app:detachedConfiguration2'
  One artifact failed verification: aapt2-8.5.1-11315950-osx.jar
```

## Why

Two independent gaps in `gradle/verification-metadata.xml`:

1. **`aapt2` was only recorded for Linux.** The file was generated on a Linux machine,
   so only `aapt2-8.5.1-11315950-linux.jar` has a checksum. `aapt2` is resolved with a
   platform classifier, so every non-Linux contributor hits a verification failure.
   This PR adds the `osx` and `windows` binaries as well.

2. **`junit-bom` was only recorded as `.pom`.** Gradle prefers Gradle Module Metadata
   and requests `junit-bom-5.9.x.module`, which was never recorded, so the buildscript
   classpath failed verification.

## Provenance of the new checksums

Every value was taken from the `.sha256` file published next to the artifact on
`dl.google.com` / `repo1.maven.org`, and cross-checked against a fresh download and
against the local Gradle cache. As a sanity check, the existing `-linux.jar` entry in
the repo matches its published `.sha256` too, confirming the same method produced the
original file.

The new entries are marked with an `origin` of `Published sha256 on ...` to distinguish
them from the trust-on-first-use entries Gradle generates.

## Verification

`./gradlew clean assembleDebug test` now passes with **strict** dependency verification
and no extra flags. Previously this required `--dependency-verification=lenient`.


peterwgnd (approved):
lgtm

PR #626, opened 2026-09-15
build: upgrade AGP to 9.4.0 and Gradle to 9.7.1

fortuna:
## What

Upgrades the Android Gradle Plugin from 8.5.1 to 9.4.0 (latest stable) and Gradle from
8.7 to 9.7.1.

## Why

AGP 8.5.1 was only tested up to `compileSdk = 34`. Since #623 compiles against 36, every
build printed:

```
WARNING: We recommend using a newer Android Gradle plugin to use compileSdk = 36
This Android Gradle plugin (8.5.1) was tested up to compileSdk = 34.
```

AGP 9.4.0 supports compileSdk 36 natively, so the warning is gone.

## What AGP 9 broke, and how it was fixed

| Breakage | Fix |
|---|---|
| `perf-plugin:1.4.1` uses `com.android.build.api.transform.Transform` and `applicationVariants`, both removed in AGP 9. The plugin failed to load. | Upgraded the Firebase plugins: google-services `4.4.1 -> 4.5.0`, perf-plugin `1.4.1 -> 2.0.2`, crashlytics-gradle `2.8.1 -> 3.0.8`. |
| `android.defaults.buildfeatures.buildconfig` was removed in AGP 9, but `LogWrapper` reads `BuildConfig.DEBUG`. | `buildFeatures { buildConfig = true }` in `app/build.gradle`. |
| `resValues` is no longer implicitly enabled, but both build types declare `FIREBASE_ENABLED`. | `buildFeatures { resValues = true }`. |
| `compileGoBackend` read `android.sdkDirectory`, removed in AGP 9. | Use the `androidComponents.sdkComponents.sdkDirectory` provider. |
| `android.nonFinalResIds=false` is no longer honored, so the `switch` statements over `R.id` in `MainActivity` stopped compiling (`constant expression required`). | Converted them to `if` / `else if` chains, matching the `id == R.id.frame_main` check that already existed a few lines below. Dropped the dead flag. |

## Regenerated lock and verification files

`buildscript-gradle.lockfile` and `gradle/verification-metadata.xml` were regenerated for
the new plugin graph.

> [!IMPORTANT]
> Regenerating on macOS records only the host `aapt2` binary, which would have
> reintroduced exactly the bug #625 fixes, just with the platforms swapped. The
> `aapt2-9.4.0-15978811` `linux` and `windows` checksums were therefore added by hand
> from the published `.sha256` files. As a cross-check, the `osx` checksum that Gradle
> generated matches its published `.sha256` exactly.

## Verification

- `./gradlew clean assembleDebug test` passes with **strict** dependency verification.

## Not addressed here

The newer lint bundled with AGP 9 reports two additional pre-existing errors that are
unrelated to the build tooling:

- `GestureBackNavigation` — `MainActivity.onBackPressed()` is not called under
  predictive back. This is a **real bug** affecting in-app back navigation, but fixing
  it properly means migrating to `OnBackPressedDispatcher`, which needs AppCompat newer
  than the pinned 1.4.0 (it resolves `androidx.activity:activity:1.2.4`, too old to
  bridge the platform callback). Worth its own PR.
- `NewApi` — `IntraTileService` requires API 24 while `minSdk` is 21. Guarded at
  runtime by the system, so this is a lint annotation gap.

`android.enableJetifier` and `android.nonTransitiveRClass` are deprecated and will be
removed in AGP 10; both are left as TODOs since migrating them needs a separate audit.


peterwgnd (approved):
lgtm

PR #627, opened 2026-09-15
build: remove deprecated enableJetifier and nonTransitiveRClass flags

fortuna:
## What

Removes the two deprecated AGP flags left behind by #626. Both are removed in AGP 10.

```diff
-android.enableJetifier=true
 android.useAndroidX=true

-# Keep legacy AGP behavior after upgrading to AGP 8
-android.nonTransitiveRClass=false
```

## `enableJetifier`

Jetifier rewrites legacy `com.android.support` artifacts to AndroidX at build time. The
dependency graph contains **zero** `com.android.support` artifacts, so this was pure
build overhead on every dependency.

## `nonTransitiveRClass`

With transitive R classes, `app.intra.R` re-exports every resource from every
dependency. Turning the flag off (the modern default) makes `app.intra.R` contain only
the app's own resources, which shrinks the generated class and makes resource ownership
explicit.

Exactly one reference needed updating. `up` and `home` in the drawer navigation handler
are declared by **AppCompat**, not by this app:

```diff
-        if (itemId == R.id.up || itemId == R.id.home) {
+        // "up" and "home" are declared by AppCompat, not by this app, so with
+        // non-transitive R classes they are not part of app.intra.R.
+        if (itemId == androidx.appcompat.R.id.up || itemId == androidx.appcompat.R.id.home) {
```

I confirmed the ownership by scanning the `R.txt` of every AAR in the dependency graph:
`up` and `home` come from `appcompat`, re-exported transitively by material,
constraintlayout, preference and firebase-perf.

> [!NOTE]
> `drawer_menu.xml` refers to `@id/home` and is deliberately left alone. Non-transitive
> R changes only the generated Java R classes; XML resource resolution goes through the
> merged resource table and is unaffected.

## Verification

- `./gradlew clean assembleDebug test` passes with strict dependency verification.
- Lint reports the same five pre-existing errors as before, so no regression.


peterwgnd (approved):
lgtm

PR #628, opened 2026-09-15
fix: restore in-app back navigation under predictive back

fortuna:
## What

Fixes in-app back navigation, which is **broken on `master` right now**, and resolves the
`GestureBackNavigation` lint error.

> [!IMPORTANT]
> This is a follow-up to #623 (already merged). That PR bumped `targetSdk` to 36, which is
> what introduced this regression. This PR is based directly on `master` and is
> independent of the AGP stack (#625/#626/#627), so it can merge on its own.

## The bug

`MainActivity` overrode `onBackPressed()` so that back returns to the home view from the
settings and info screens:

```java
  @Override
  public void onBackPressed() {
    if (isShowingHomeView()) {
      super.onBackPressed();
      return;
    }
    chooseView(R.id.frame_main);
  }
```

Apps targeting **SDK 36** can no longer opt out of predictive back, and the platform
routes back gestures through `OnBackPressedDispatcher` instead of calling
`onBackPressed()`. So pressing back from Settings **exits the app** rather than returning
to the home view.

This is specific to `targetSdk` 36 — it is not a pre-existing bug. Running `lintDebug`
across `targetSdk` values on an otherwise identical tree:

| targetSdk | total lint errors | `GestureBackNavigation` |
| --- | --- | --- |
| 33 | 3 | not reported |
| 34 | 4 | not reported |
| 35 | 4 | not reported |
| 36 | 5 | **reported** |

targetSdk 35 (Android 15) turned on predictive back *animations* but kept the legacy
`onBackPressed()` path working. targetSdk 36 (Android 16) removes that compatibility path.

## The fix

Register an `OnBackPressedCallback` on the dispatcher instead, enabled only while a
non-home view is showing:

```java
  private final OnBackPressedCallback backToHomeCallback = new OnBackPressedCallback(false) {
    @Override
    public void handleOnBackPressed() {
      chooseView(R.id.frame_main);
    }
  };
```

`chooseView()` is the single funnel for every view change, so it is the natural place to
keep the enabled state in sync. It sits right next to the existing up-arrow logic, which
already encodes the same "am I on the home view" condition:

```java
    backToHomeCallback.setEnabled(id != R.id.frame_main);
```

When disabled, back falls through to the system and leaves the app, preserving the
original behavior. `isShowingHomeView()` has no remaining callers and is removed.

## Why AppCompat moves 1.4.0 -> 1.7.1

Bridging the platform's `OnBackInvokedCallback` into `OnBackPressedDispatcher` needs
`androidx.activity` 1.6+. The pinned appcompat 1.4.0 resolved `activity:1.2.4`, which
predates predictive back entirely. appcompat 1.7.1 pulls `activity:1.8.0`.

> [!NOTE]
> appcompat **1.8.0** is newer, but requires raising `minSdk` to 23:
> ```
> uses-sdk:minSdkVersion 21 cannot be smaller than version 23 declared in library
> [androidx.appcompat:appcompat-resources:1.8.0]
> ```
> 1.7.1 is used deliberately so that `minSdk` stays at 21 and Android 5.x users are not
> dropped. For a censorship-circumvention tool, old devices are worth keeping.

`gradle/verification-metadata.xml` is updated with the new transitive dependencies
(activity 1.8.0, lifecycle 2.6.x, core 1.13.0, emoji2, savedstate, profileinstaller, ...).

## Verification

- `./gradlew clean assembleDebug test` passes.
- `GestureBackNavigation` is gone. The remaining lint errors are the 3 pre-existing ones
  (`StartActivityAndCollapseDeprecated`, `RestrictedApi`, `QueryAllPackagesPermission`).

Exercised on a **Pixel 8 emulator running API 36** (the configuration where the bug
reproduces), driving the UI over `adb` + `uiautomator`:

| Check | Result |
| --- | --- |
| Drawer -> Settings shows the settings view | pass |
| Back from Settings stays in the app | pass |
| Back from Settings returns to the home view | pass |
| Tapping a stat box shows the info view | pass |
| Back from the info page stays in the app | pass |
| Back from the info page returns to the home view | pass |
| Back from the home view leaves the app | pass |
| No crashes in logcat | pass |

Both `chooseView()` entry points are covered: the settings view and the info view. Back
from the home view still exits, so the pre-existing behavior is preserved.


jyyi1 (approved):
LGTM, thanks!

